PR #118, opened 2015-01-30
Possible translation page 404

onedeko:
I added the ability to translate the 404 page.


daveismyname:
i'd accept the 404 changes but i'm not renaming error to alert.

PR #122, opened 2015-02-26
Develop

pixelspezi:


daveismyname:
thanks but I think returning htmlspecialchars by default is not always needed, best to return the data as is then let the developers change it on a as needed basis.

PR #127, opened 2015-03-23
Cleaned up router a little bit

Volter9:
I'm back! :stuck_out_tongue: 
Decided to make a little clean up contribution. Also, just wonder, what this chunk of code is doing:
```php
$char_position = strpos($controller,'&');
if ($char_position > 0 ) {
    $ctp = explode('&', $controller);
    $controller = $ctp[0];
}

$char_position2 = strpos($method,'&');
if ($char_position2 > 0 ) {
    $ctp = explode('&', $method);
    $method = $ctp[0];
}

if ($args != null) {
    $char_position3 = strpos($args[0],'&');
    if ($char_position3 > 0 ) {
        $ctp = explode('&', $yes);
        $args[0] = $ctp[0];
    }
}
```
Could anybody explain what's it doing? Searching for query ($_GET) parameters or something like that and cleaning it up?


QsmaPL:
I think:

This check if position of char "&" for controller/method/args is bigger than 0.

for example if position of char "&" in $controller is bigger than 0 next explode $controller and get first part.

and all new values ($method, $controller, $args) are displayed on screen from controller and new method / args ;p

In my opinion router is too big, i future i want create other router, simpler and more flexible. 
Maybe we could project new algorithms for router and create it?

Volter9:
> In my opinion router is too big

It's true, it does too much stuff. 

Recently, I created two versions of controller using OOP techniques, first one is simple one, it is about 120 lines long, there's 5 classes, and it's working like autoDispatch: routing [component](https://bitbucket.org/miraplanet/mira-planet/src/97d3ccde7927c0a823484dd7c42e6a9edfc60be6/framework/routing/?at=volter) and [example](https://bitbucket.org/miraplanet/mira-planet/src/97d3ccde7927c0a823484dd7c42e6a9edfc60be6/framework/core/App.php?at=volter#cl-20) (there's [invoke](https://bitbucket.org/miraplanet/mira-planet/src/97d3ccde7927c0a823484dd7c42e6a9edfc60be6/site/index.php?at=volter#cl-22), sorry about the Russian comments :laughing: ).

There's also another one that I developed, it's much complex than previous one. There it [is](https://github.com/Volter9/Trial/tree/master/Trial/Routing). It's much bigger and works like this router without auto dispatch, it's very complex, and it has about 12-14 classes with total of 500 lines.
You can check them both, however, the complex one has a dependency upon DIC (which is ironically, because DIC suppose to remove the dependency or at least loosen it up :laughing: ).

Volter9:
Sorry for littering with small commits, forgot to change everything back. Basically, I figured out why this code was needed. This code was needed to clear query strings (i.e. $_GET arguments), I came up with simpler way:

1. If requested URI contains `&`, than remove everything after `&` in URI. 

That's it. Now it should look a little bit cleaner.

daveismyname:
Is this the updated version now? one think I've noticed trying it our when using domain/controllers/param calls any params passed where in an array now they are separate variables, whilst it does make sense I'm not sure if this should be changed now.

I use a lot of domain/controller calls and use the array changing the router to what you have done means I'll need to update my controller methods getting on for around 40ish now. I would suspect this may be the case for others too.

I don't want to break existing code for people, thoughts?

Volter9:
@daveismyname are you talking about array as first argument? If so, I probably broke it by mistake and I'll return back after you'll reply. Ok?

daveismyname:
Yes I think that's it.

Volter9:
@daveismyname returned it, is now everything is fine?

daveismyname:
yes looks to be working great thanks.

PR #130, opened 2015-04-21
Suggestion password helper

QsmaPL:
In the recent time I created helper to the password hashing and verify.
My class is smaller and more legible than current. Could somebody look at it and test?

repo: https://github.com/QsmaPL/Simple-PHP-Password-hasher

It is possible so that it is to add it to SMVCF and to replace old class? :)

### DIFF
--- a/app/helpers/password.php
+++ b/app/helpers/password.php
@@ -1,224 +1,86 @@
-<?php namespace helpers;  
+<?php namespace helpers;
+
 /**
- * A Compatibility library with PHP 5.5's simplified password hashing API.
- *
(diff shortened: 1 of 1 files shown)


daveismyname:
This looks great but I won't be replacing the password class, their is more content and complexity on the password class granted but it's what's recommended to use the same functionally as the built in password functions for php 5.5.

This is for users that are on a version less then 5.5

QsmaPL:
Okey, pls update current class to latest version https://github.com/ircmaxell/password_compat/blob/master/lib/password.php ;)

daveismyname:
yes good call I will look at updating it this evening.

PR #140, opened 2015-05-21
Added german language and fix button label

nekomajin:


daveismyname:
var should no not start with a _ as per PSR recommendations as such $_array should be $array

nekomajin:
Ok.So how can we solve the problem?

daveismyname:
what is the problem? push a PR with the language addition and I'll accept it.

PR #152, opened 2015-06-06
Fixed the wrong translation (Maybe google translator?) to right german

BlueFreakLP:
The translation was not right for the german language, so i fixed it.


nekomajin:
I wanted to stick to the english original. Was no Google Translator. Thanks anyway :+1:

PR #160, opened 2015-07-14
Update Csrf.php & Create Request.php

OmarElGabry:
The current implementation of CSRF tokens can be enhanced:

1 . The current implementation is per-form; means if you have multiple opened pages from your website, and each page has a form, you will ending up having different tokens in each form.

+ This could result in invalid CSRF token while the form is submitted from legitimate user(not faked).
**Please** check this [thread](http://stackoverflow.com/questions/10466241/new-csrf-token-per-request-or-not?lq=1) 
+ Adding expiry date is better for making it more secured.

2 . It's painful to re-write the following snippet in every view page.
```php
    use Helpers\Session;
    $token = trim($data['token']);
    Session::set('token', $token);
```
Instead this should work 
```
    <input type="hidden" name="csrf_token" value="<?= $data['csrf_token']; ?>" />
```

3 . As well, validating the CSRF token could be encapsulated in a method. You may want to validate if the HTTP Method is POST before calling this method.
```php
    if (!Csrf::isTokenValid()) {
    }
```

### DIFF
--- a/app/Helpers/Csrf.php
+++ b/app/Helpers/Csrf.php
@@ -2,6 +2,7 @@
 
 namespace Helpers;
 
+use Helpers\Session;
 /**
  * Cross Site Request Forgery helper
  *
@@ -14,34 +15,54 @@ namespace Helpers;
  * At the top of the controller where the other "use" statements are place:
  * use Helpers\Csrf;
  * 
- * Just prior to rendering the view for adding or editing data create the token:
- * $data['token'] = Csrf::makeToken();
+ * Just prior to rendering the view for adding or editing data create the CSRF token:
+ * $data['csrf_token'] = Csrf::makeToken();
  * $this->view->renderTemplate('header', $data);
  * $this->view->render('pet/edit', $data, $error); // as an example
  * $this->view->renderTemplate('footer', $data);
+ *
+ * At the bottom of your form, before the submit button put:
+ * <input type="hidden" name="csrf_token" value="<?= $data['csrf_token']; ?>" />
  * 
- * In the add or edit view (form) at top of page place:
- * <?php
- * use \Helpers\Session;
- * $token = trim($data['token']);
- * Session::set('token', $token);
- * ?>
- * 
- * Towards the bottom of your add or edit form (same form) before the submit button put:
- * <input type="hidden" name="token" value="<?php echo $token; ?>" />
- * 
- * These lines need to be placed in the controller action for the update or add method:
- * if ($_POST['token'] != Session::get('token')) {
- *     Url::redirect('admin/login'); // or wherever you want to redirect to.
+ * These lines need to be placed in the controller action to validate CSRF token submitted with the form:
+ * if (!Csrf::isTokenValid()) {
+ *      Url::redirect('admin/login'); // or wherever you want to redirect to.
  *    }
  * And that's all 
  */
 
 class Csrf {
 
+    /**
+     * get CSRF token and generate a new one if expired
+     *
+     * @access public
+     * @static static method
+     * @return string
+     */
     public static function makeToken() {
-        $csrf_token = md5(uniqid(rand(), TRUE));
-        return $csrf_token;
+
+        $max_time    = 60 * 60 * 24; // token is valid for 1 day
+        $stored_time = Session::get('csrf_token');
+        $csrf_token  = Session::get('csrf_token_time');
+
+        if($max_time + $stored_time <= time() || empty($csrf_token)){
+            Session::set('csrf_token', md5(uniqid(rand(), true)));
+            Session::set('csrf_token_time', time());
+        }
+
+        return Session::get('csrf_token');
+    }
+    
+    /**
+     * checks if CSRF token in session is same as in the form submitted
+     *
+     * @access public
+     * @static static method
+     * @return bool
+     */
+    public static function isTokenValid(){
+        return $_POST['csrf_token'] === Session::get('csrf_token');
     }
 
 }
--- a/app/Helpers/Request.php
+++ b/app/Helpers/Request.php
@@ -1,0 +1,77 @@
+<?php
+namespace Helpers;
+
+/*
+ * Request Class
+ *
+ * It contains the request information and provide methods to fetch request body
+ */
+class Request{
+
+    /**
+     * safer and better access to $_POST
+     *
+     * @param  string   $key
+     * @static static method
+     * @return mixed
+     */
+    public static function post($key){
+        return array_key_exists($key, $_POST)? $_POST[$key]: null;
+    }
+
+    /**
+     * safer and better access to $_FILES
+     *
+     * @param  string   $key
+     * @static static method
+     * @return mixed
+     */
+    public static function files($key){
+        return array_key_exists($key, $_FILES)? $_FILES[$key]: null;
+    }
+
+    /**
+     * safer and better access to $_GET
+     *
+     * @param  string   $key
+     * @static static method
+     * @return mixed
+     */
+    public static function query($key){
+        return array_key_exists($key, $_GET)? $_GET[$key]: null;
+    }
+
+    /**
+     * detect if request is Ajax
+     *
+     * @static static method
+     * @return boolean
+     */
+    public static function isAjax(){
+        if(!empty($_SERVER['HTTP_X_REQUESTED_WITH'])){
+            return strtolower($_SERVER['HTTP_X_REQUESTED_WITH']) === 'xmlhttprequest';
+        }
+        return false;
+    }
+
+    /**
+     * detect if request is POST request
+     *
+     * @static static method
+     * @return boolean
+     */
+    public static function isPost(){
+        return $_SERVER["REQUEST_METHOD"] === "POST";
+    }
+
+    /**
+     * detect if request is GET request
+     *
+     * @static static method
+     * @return boolean
+     */
+    public static function isGet(){
+        return $_SERVER["REQUEST_METHOD"] === "GET";
+    }
+
+}


OmarElGabry:
For Request.php:

Abstracting the access to super globals is much cleaner way. Please take a look at this [thread](http://stackoverflow.com/questions/13359818/what-is-the-right-way-to-handle-post-data-in-mvc)

jimgwhit:
You still need the token put in a post variable and a session variable this is how It gets verified they have to match. Read http://shiflett.org/articles/cross-site-request-forgeries

OmarElGabry:
**Please read the the description above, my comments and the code well.** 

1. The token will be in hidden field as before

2. makeToken() is responsible for adding it to the session. This will prevent me from adding it to the session every time in my view at the top.

3. They are verified the same as before, but again, instead of re-writing 
```php
    if ($_POST['csrf_token'] != Session::get('csrf_token')) {
    }
```
This snippet will be encapsulated inside a method. So, if you want to change the way you verify, you will only need to change this method without affecting the rest of your code.
```php
    if (!Csrf::isTokenValid()) {
    }
```

Everything is explained and commented.

jimgwhit:
My whole idea was to actually have a brand new token each and every time you post not just have one stored for a while it's just more secure that way.

jimgwhit:
Also it still needs to be trimmed don't ask me why but in regular PHP sessions it works but in the session helper it does not work without the trim command.

OmarElGabry:
I don't know why you want to keep arguing. if you want people to fork, and contribute, then give a constructive feedback. Ahaaa I see that you are the one who created CSRF.php file, That's why you are arguing ? :D . Your name as author in CSRF.php won't be replaced, you are the author, happy now?

Ok, here again:

1 . Generating a new token for each form could report a false CSRF attack. That's likely to happen when you open more than one form at the same time, because one of them will have a token that's one same as the session. For more details Please check this [thread](http://stackoverflow.com/questions/10466241/new-csrf-token-per-request-or-not?lq=1) 

So, what shall we do? Simply if you created per-session token with expiry time is enough to keep it secured. Try to dig into websites like [Linkedin](Linkedin.com) and see how those people are implementing the CSRF tokens.

2 . Your request sounds like, "Hey man I don't care about what you are doing". Everything is commented and clear. Ok, Even though, here you are:

```php
    // get value from $_POST
    $username = Request::post('username');

    // get value from $_GET
    $username = Request::query('id');

    // get value from $_FILES
    $username = Request::files('file');

   // is the request was POST ?
   if(Request::isPost()){
   }

```

3 . I don't know why you need to trim the token, the value returned from ``` md5(uniqid(rand(), true)) ``` doesn't need to be trimmed.

I don't mind to keep this argument as long you will give me a constructive feedback.

daveismyname:
I like this improvement, makes the implementation a smoother process, using a single method to generate the token and return and stores it in a session is nice.

Then processing the form becomes easier too.

The new request helper is nice too alothugh when accessing a $_GET why is the method called query, get seems a more obvious choice and fits in with the other conventions.

Any valid objections to using this updated helper?

OmarElGabry:
It's called ```query()``` referring to querystring arguments(URL query data) in $_GET. I was thinking about calling it ```get()``` instead. But, This also reveals ambiguity because anyone may think of it as getters and setters instead of accessing $_GET.

Anyways, I saw the method name ```query()``` in CakePHP [query()](http://api.cakephp.org/3.0/source-class-Cake.Network.Request.html#_query).

daveismyname:
That's a good point personally I would prefer get seems more natural post is using post, files is using files..ect

calling it query or get at the end of the day is not going to make a huge difference.

OmarElGabry:
Yes :+1:

jimgwhit:
In my setup session helper would not work with it unless it is trimmed and I do not know why I tested it using a regular PHP session and it worked without trim maybe Dave can shed some light on this. Also if this new CSRF is implemented please give detailed instructions on usage because I have a production site that I try to keep up to date thanks.

daveismyname:
the implementation looks close to how it was:
    
Controller:

Validation process:

    if (!Csrf::isTokenValid()) 
    {
        Url::redirect('admin/login'); // or wherever you want to redirect to.
    }

Set data keys

    $data['csrf_token'] = Csrf::makeToken();

View:

    <input type="hidden" name="csrf_token" value="<?= $data['csrf_token']; ?>" />

Have I missed anything?

OmarElGabry:
Yes, Exactly. I've written instruction at the beginning in [CSRF.php](https://github.com/OmarElGabry/framework/commit/43a8f3de30435ffe52a30780f90d216c2f000821)

```php
/**
 * Instructions:
 * At the top of the controller where the other "use" statements are place:
 * use Helpers\Csrf;
 * 
 * Just prior to rendering the view for adding or editing data create the CSRF token:
 * $data['csrf_token'] = Csrf::makeToken();
 * $this->view->renderTemplate('header', $data);
 * $this->view->render('pet/edit', $data, $error); // as an example
 * $this->view->renderTemplate('footer', $data);
 *
 * At the bottom of your form, before the submit button put:
 * <input type="hidden" name="csrf_token" value="<?= $data['csrf_token']; ?>" />
 * 
 * These lines need to be placed in the controller action to validate CSRF token submitted with the form:
 * if (!Csrf::isTokenValid()) {
 *      Url::redirect('admin/login'); // or wherever you want to redirect to.
 *    }
 * And that's all 
 */
```

For trimming the values, I tested as well, It don't see any need to trim values. The current implementation:
```php
function makeToken() {
	$csrf_token = md5(uniqid(rand(), TRUE));
	return $csrf_token;
}

var_dump(makeToken());
```

If anyone inspected any issue with the current implementation(the token still needs to be trimmed), please provide us with your test function, screenshot, or anything else.

jimgwhit:
On your request pull request you are still just using post and get etc, this pull request would make more sense if you had the PHP filter_input functions integrated. See http://php.net/manual/en/function.filter-input.php

jimgwhit:
In your var dump you're using regular PHP session and I don't see the session helper try it using the session helper. Again I had to use trim for it to work properly using session helper. And really no arguments here I just want to make sure everything works real good.

jimgwhit:
Forgot to mention but an example in the request would be stripping tags I have a custom function not included in the framework that strips tags. In the request class give user the choice of what to clean and what not to clean, I think it would just be a good idea. On tablet now will try to give an example when on PC.

jimgwhit:
On pc now, as an example I have this utility class I call Clnsantize:  
~~~
<?php

namespace Controllers;

class Clnsantize
{

    private $rvalue;
    private $rchk = "0";
    private $tonl;

    public function fixValue($rvalue)
    {
        $this->rvalue = empty($rvalue) && !is_numeric($rvalue) ? NULL : trim(strip_tags($rvalue));
        return $this->rvalue;
    }

    public function fixCheck($rchk)
    {
        echo 'in func';
        $this->rchk = ($rchk == 'on' || $rchk == '1' || strlen($rchk) > 0 ? '1' : '0');
        //$this->rchk = $rchk;
        return $this->rchk;
    }

    public function toUpper($rchk)
    {
        $this->rchk = strtoupper($rchk);
        return $this->rchk;
    }

    public function fixInt($tonl)
    {
        $this->tonl = (is_null($tonl) || empty($tonl) || strlen($tonl) < 1 ? NULL : $tonl);
        //if( is_null($tonl) || empty($tonl) || strlen($tonl) < 1 )
        //{ 
        //$tonl = NULL;
        //} 



        return $this->tonl;
    }

}

//end class

~~~
Could you possible implement something like this in your request class?  An example is dates, I hate when a date field shows 0000-00-00.  I make that a null if no date is entered.via the fixInt function.  I just call it fixInt, it's for any field numeric field you want null that someone left blank.  Again this is just example of how you could add more stuff to the request class.  I hope you will.  
Not the fixcheck one it was an experiment gone bad.

OmarElGabry:
Ok,

>On your request pull request you are still just using post and get etc, this pull request would make more sense if you had the PHP filter_input functions integrated. See http://php.net/manual/en/function.filter-input.php

PHP filter_input are used to get a value for a key in super globals(GET, POST, ...), and filters it. Request helper has a responsibility, and it should have **only** one responsibility, which is *"contains the request information and provide methods to fetch request body".*

If you want to filter, you should have something else called for example "Utility" that takes care of filtering the values.

Request shouldn't take care of cleaning the values. It's out of it's responsibility. What you are suggesting is very bad design practice. You are trying to build what I call "God Class"; the class that does everything.

Why you should avoid this? Because the more responsibilities you have, the more possible bugs, harder to fix and modify, and can affect on other components in your application.

 >In your var dump you're using regular PHP session and I don't see the session helper try it using the session helper.

What's your problem? the code I've used is same as in session helper, just something quick, I used generic functions instead of installing the whole framework. But, the code is the same.

> Again I had to use trim for it to work properly using session helper. And really no arguments here I just want to make sure everything works real good.

You want to make sure everything works good? Then, test your code, show us your results, and what you have got. There is no place for feelings, assumptions, or hypothesis. You want to make sure of something, then go and test it under different circumstances and compare the results you got with the expected one.

Volter9:
Sorry for offtop :smiley:
Keep going:

michael popcorn

OmarElGabry:
:smile: I see that you are one of the contributors

jimgwhit:
Please read https://www.inanimatt.com/php-input-filtering.html it talks about how the Symfony HttpFoundation Component filters. This is what I AM referring to do some basic filtering in the class otherwise you're just getting a post value and returning it no different than using a post in the controller. And you need to reread the PHP manual filter input replaces post as a preferred technique. Of course you can use the class if you wish, I see no reason for the class if it only returns exactly what's been sent to it.

jimgwhit:
P.S. Normally a class does something to the data, I.e.,  filters, sanitizes, something you are basically sending foo to the class and returning foo exactly as is no changes no checking no difference. To me a class should have some purpose. I do not want simple MVC to turn into a monster like Laravel, for specialized classes I feel people should write their own and not add it to a framework like my clean sanitize class above, I did not put in a pull request on that. Let's keep this framework simple and not bogged down with a bunch of stuff. However I am all for your class if it had basic sanitizing but let the user choose what to sanitize and how to sanitize it, by having different filtering and/or sanitizing methods in the class.

mherrm:
I would like to see such class as a part of the framework because its a basic thing every web application does.

jimgwhit:
@Volter9  you should have put on the animated gif of this much funnier. See https://laracasts.com/discuss/channels/laravel/laravel-is-a-giant-waste-of-time/?page=1 for some funny gifs

OmarElGabry:
> Please read https://www.inanimatt.com/php-input-filtering.html it talks about how the Symfony HttpFoundation Component filters. This is what I AM referring to do some basic filtering in the class 

Not because symphony takes a specific approach, you should do so. If symphony does, a lot of other frameworks don't. Even though, they are implementing in a separate method, and not inside access methods for *$_GET*, *$_POST*, *$_FILES*.

If you need to add filtration method, you keep it objective as much as you can. Don't get beyond the scope. Thus, Request class will and should have only **one** responsibility. If you need to get further, read about [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle)

> "states that every class should have responsibility over a single part of the functionality provided by the software, and that responsibility should be entirely encapsulated by the class"

Things like striping tags, formatting dates,  Convert characters to HTML entities, url encode, json encode, and so on should be part of your View(View Helpers), and you would probably use them right before displaying the values.

> otherwise you're just getting a post value and returning it no different than using a post in the controller. 

The idea of Request is it prevents you from direct access to super globals, instead you access them through convenient method. In addition, There are some other useful methods like ```isPost()```, ```isAjax()```, and so forth.
If you have ever used static analyzers, you would probably noticed many complaints because it's discouraged to access super global directly.

> I do not want simple MVC to turn into a monster like Laravel, for specialized classes I feel people should write their own and not add it to a framework like my clean sanitize class above, I did not put in a pull request on that. Let's keep this framework simple and not bogged down with a bunch of stuff.

I don't see any complexity here, It's much cleaner and convenient to follow this snippet. And for all people in the house, tell me which is better?

```php
    // get value usingRequest::post()
    $username = Request::post('username');
   
    // get value using $_POST directly
    $username = $_POST['username'];
```

See, It's not about "Who is going to win the race, you or me?". You don't pay for me, this is not something you will pay for. I sent a pull request, and It has been 4 days arguing about some issues without giving any constructive feedback.

You started arguing about the CSRF class, and how it works, Although everything was well explained and documented. Then, you turned on arguing about the token and how it should work. Then, another one about Request, and how it works, again everything was clear. Moving on, you started the argument about trimming the values(without giving any test results).

Then, you got another an idea about having sanitizing and filtration inside Request class. You started the argument by trying to convince that Request class should also sanitize and filter the values, like strip tags, and dates as you suggested. 

What's next?

It's not a shame if you have something wrong, and i am not an expert though, I don't know why are not accepting any feedback, and you keep try to prove what's in your mind without giving any valid objections.

I would be glad, and honor if you disagree because you really have a good understanding, and you will give me a constructive feedback that will make things clear. Other than that, Sorry man But, It's a waste of time and efforts.

OmarElGabry:
I would be also very happy if any contributors get evolved in this conversation and explain their thoughts.

jimgwhit:
Again its fine with me if the pull request goes through my whole point is you sending foo to a class and getting the exact same thing back that you are sending. A simple MVC user may as well just use the PHP  filter input functions which if an argument is left blank it doesn't filter anything and that is an approved industrial strength accepted practice. And you're not accessing super globals directly that way. Now I have referred you to the PHP manual and PHP approved accepted techniques, that is about as constructive as I can get sorry. And I have no issue with your CSRF improvement I just haven't had a chance to test yet on my computer but it sounds good. All I know is in my setup I had to trim a token for it to work with sessionhelper but using regular PHP sessions trimming was not necessary please I do not want to have to repeat that again on my setup it was that way why I do not know. I develop on windows in netbeans I do not know if that makes any difference or not.

Volter9:
@OmarElGarby I like those new improvements, the CSRF and Request utility, however, I a little bit concerned about urs static methods which is leading to tight coupling. For beginners, it's easier to use static classes, but for flexibility, it will be better if it be an instance class.

jimgwhit:
I am going to try to test CSRF helper tonight if I get a chance and tests that trimming vs non trimming and see if it works if it doesn't work on mine I have no idea why. Will post results.

OmarElGabry:
@Volter9  Yes this is true, I wanted to do so. But, When I went thought the helper classes, and i found they use static methods, So, I kept them as static. 

But, I would recommend to access methods through instances Vs static methods. Not only in CSRF, Request, but also in other helper classes. This modification should take place across all helper classes.

OmarElGabry:
@jimgwhit See, do whatever you want to do :smile: 

You can Merge the code, then add your own modifications. If you want to add filtration arguments to Request, It should be done by you(It won't require huge efforts), because you are the one who believe in, It's your idea man, you are the author, Remember?

jimgwhit:
I tested and all ok.  Strange, I retested mine also without trim and now works.  I think I had php 5.4.-- and now 5.6.--.  Don't know why trim was needed before.  But yours works great. I had something similar happen once when trying to verify a password, had to trim. Maybe a fluck in the ver of php at the time.

jimgwhit:
@OmarElGabry I can't merge, I don't have that access, Dave will have to merge.

daveismyname:
>But, I would recommend to access methods through instances Vs static methods. Not only in CSRF, Request, but also in other helper classes. This modification should take place across all(as long as it's applicable) helper classes.

Most helpers have static setup due to them being individual actions the helpers that tie in to other actions are instances.

I'm curious which helpers do you see been better as instances?

OmarElGabry:
When I run into a situation where i want to determine whether to use static Vs instance method. I ask myself, *"Is the method tied to the object instantiated form that class; In other words, is the method represents a behavior/functionality for each object, Or the method is independent on the instantiated object?"*

As an example, ```get()``` method in [Database](https://github.com/simple-mvc-framework/framework/blob/master/app/Helpers/Database.php) Class is static, while ```select()```, ```insert()```, ...etc. are instance methods.

In some situations, for me, they could be static and instance methods. For example, one could say [Form](https://github.com/simple-mvc-framework/framework/blob/master/app/Helpers/Form.php) Class should have static methods, and another one may say the methods aren't tied to the instantiated object.

Case studies from frameworks:
+ CakePHP [Form Helper](http://book.cakephp.org/2.0/en/core-libraries/helpers/form.html) uses object to create form fields 
+ Laravel [Form Class](http://laravel.com/docs/4.2/html) uses static methods instead.

So, It's up to you, use whatever is convenient for you. The candidate helpers that could be used as instances: ```Form```, ```Request```, ```Session```, ```SimpleCurl```, ```Url```. But, if you kept them as static, this also would be logically acceptable.

Also In Core/View, You may want to convert render methods to instances, and you can use the view instance inside Core/Controller ```$this->view = new View();```

PR #167, opened 2015-08-03
Segments routes with callback

fabioassuncao:


daveismyname:
cannot accept the methods they are not psr 1 & 2 compliant, segments are already in use the changes proposed would break existing implementations.

PR #169, opened 2015-08-10
Swapped varibles $csrf_token and $stored_time

capps1994:
csrf token kept being regenerated this is now resolved

### DIFF
--- a/app/Helpers/Csrf.php
+++ b/app/Helpers/Csrf.php
@@ -43,8 +43,8 @@ class Csrf {
     public static function makeToken() {
 
         $max_time    = 60 * 60 * 24; // token is valid for 1 day
-        $stored_time = Session::get('csrf_token');
-        $csrf_token  = Session::get('csrf_token_time');
+        $csrf_token = Session::get('csrf_token');
+        $stored_time  = Session::get('csrf_token_time');
 
         if($max_time + $stored_time <= time() || empty($csrf_token)){
             Session::set('csrf_token', md5(uniqid(rand(), true)));


jimgwhit:
Dave I don't think there is an issue here I believe because the example shows short tags can you change the example in the documentation to use regular tags. But can you test first and make sure that is correct. From. 
~~~
 <input type="hidden" name="csrf_token" value="<?= $data['csrf_token']; ?>" />
~~~
To. 
~~~ 
<input type="hidden" name="csrf_token" value="<?PHP echo $data['csrf_token']; ?>" />
~~~  
I believe that makes the difference, please verify.

jimgwhit:
Also pass on to anyone regular style should be used in any examples as some servers and people don't use short tags. Thanks

capps1994:
This patch is nothing to do with the short tags. You are correct that short tags shouldn't be used, but I have tested with both short and normal tag.

The issue reside on the check to recreate the csrf.

This is a snippet of the orginal code.

Correct me if I am wrong but it does look like the csrf would get regenerated at randoms times.

```
$stored_time = Session::get('csrf_token');
$csrf_token  = Session::get('csrf_token_time');

 if($max_time + $stored_time <= time() || empty($csrf_token)){
             Session::set('csrf_token', md5(uniqid(rand(), true)));
```

Thanks

jimgwhit:
I finally had time to test: I did this:  
Right after the verification
~~~
 if (!Csrf::isTokenValid()) {
                Url::redirect('admin/login'); // or wherever you want to redirect to.
            }
~~~
I compared the session and post like:
~~~
$ttest = Session::get('csrf_token');
            echo $ttest;
            echo "<br>";
            echo $_POST[csrf_token];
~~~
And they matched perfectly:
~~~
4de7440fc232f10a0265d5ace0c6b1b7
4de7440fc232f10a0265d5ace0c6b1b7
~~~
So I still believe the problem was short php tags,  
This
~~~
<input type="hidden" name="csrf_token" value="<?= $data['csrf_token']; ?>" />
~~~
Whereas this should be used:
~~~
<input type="hidden" name="csrf_token" value="<?PHP echo $data['csrf_token']; ?>" />
~~~
It is working fine on my development server and production.

jimgwhit:
Dave can you verify I tested correctly?

jimgwhit:
capps1994 whent is it happening?  I will try to test more.

jimgwhit:
Hey, I thing I see what you mean, the vars names don't match:  
~~~
$stored_time = Session::get('csrf_token');
$csrf_token  = Session::get('csrf_token_time');
~~~
I missed that at first.  I also am swapping them and doing more test, sorry I missed that.

capps1994:
Yes that is what I mean. it's bit of a itermitent bug, where some times it will work, I was using it a whole day before I noticed that it was doing weird things to the csrf keys. I think it happens when you don't destroy the session probably (just closing the browser)

jimgwhit:
Dave please implement this.

jimgwhit:
Of course tset first if needed.

capps1994:
It's fine very easily done, thank you for testing it to make sure it was working

OmarElGabry:
@capps1994 That's right!, that's a mistake that i admit I've made. I am sorry for that, but what you are suggesting is right and should be modified as soon as possible. Thanks!

OmarElGabry:
@daveismyname Can you please merge this request as soon as possible as it's extremely important :) ? Thanks

jimgwhit:
Dave read issues #173 and #174, you&#39;ll like what volter9 and myself came up with, and his new pull request. With pull request #175 it will make the framework PHP 5.4, PHP 5.5, and PHP 7 ready. Of course you&#39;ll test everything thoroughly.

PR #175, opened 2015-08-25
Adjusted invokeObject method in router for PHP7

volter9:
Thanks to @jimgwhit for testing the code on PHP7 :smiley:

Issue #174, opened 2015-08-25
Should Core\Controller be regular class, not abstract


jimgwhit:
I know somehow the router is working, however in netbeans ide I get:
```
Abstract class \Core\Controller can not be instantiated
```
It happens here at line 126:
```
$c = new $controller;
```
The new key word isn't supposed to work with an abstract class, but yet the router is working.  
I am no expert on writing router code, so could someone shed some light on how the router is still working anyway -- thanks.

volter9:
Just ignore this warning. In the past, netbeans gave me loads of useless warnings.

The router works following way:
1. Match given route with given request url and method
2. If the route is found, execute it by creating controller instance and invoking method in the given route
3. Else, 404

jimgwhit:
I found out something about the router code.  
This code in the public static function invokeObject($callback, $matched = null, $msg = null) function works in php 5.5 and php 7.--:
```
public static function invokeObject($callback, $matched = null, $msg = null)
    {

        $last = explode('/', $callback);
        $last = end($last);

        //grab the controller name and method call
        list($controller, $method) = explode('@', $last);

        //instanitate controller with optional msg (used for errorCallback)
        $controller = new $controller($msg);

        if ($matched == null) {
            //call method
            $controller->{$method}();
        } else {
            //call method and pass in array keys as params
            call_user_func_array(array($controller, $method), $matched);
        }
    }
```
But the original code that's present now does not work in php 7.  
I will probably do a pull request, any thoughts.  Changing this one function in router makes the framework work in php 5.--- and php 7.---  
This is the only function that would have to be changed in the router.

jimgwhit:
Above code came from @pandory-network, another SMVC contributor, I just never thought to  
try it the other way around, using it in php 5.5 and php 7.---

jimgwhit:
Oh,  @pandory-networkline said that line 92 in router causes it not to work in php 7.  
Why doesn't
```
 $controller->$segments[1]();
```
not work in php 7?

volter9:
Because it's treated as `$controller->{$segments[1]}();`.
Read about the changes in code parsing in PHP7.

jimgwhit:
Thanks for reply, @volter9 how would you write line 92 to work in both versions of php?

volter9:
I'm not really sure. I installed php7, but I'm too lazy to test it. Therefore, I installed only php7 for CLI, not really sure how to bind it to apache. Did you tried the code I suggested from issue #173 with php7?

[...]

PR #177, opened 2015-08-29
Multiple features & upgrades

ennerperez:
Uppercase folders in app folder
APPSTATE for offline/online environments
Separated routes from index
Language resx call style
Logger line fix for more readability
Include images in Assets
Included MSSQL for Windows Servers
Included plain text based GeoName

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 app/Core/Config.php
 .DS_Store
+*.project
+/aspnet_client
--- a/app/Core/Config.example.php
+++ b/app/Core/Config.example.php
@@ -1,4 +1,4 @@
-<?php
+﻿<?php
 namespace Core;
 
 use Helpers\Session;
@@ -12,6 +12,8 @@ use Helpers\Session;
  * @version 2.2
  * @date June 27, 2014
  * @date updated May 18 2015
+ * @author Enner Pérez - ennerperez@gmail.com
+ * @date updated Jun 29 2015
  */
 class Config
 {
@@ -32,6 +34,7 @@ class Config
 
         //set a default language
         define('LANGUAGE_CODE', 'en');
+        setlocale ( LC_TIME, "en_US" );
 
         //database details ONLY NEEDED IF USING A DATABASE
         define('DB_TYPE', 'mysql');
--- a/app/Core/Language.php
+++ b/app/Core/Language.php
@@ -10,6 +10,8 @@ use Core\Error;
  * @version 2.2
  * @date November 18, 2014
  * @date updated May 18 2015
+ * @author Enner Pérez - ennerperez@gmail.com
+ * @date updated Jun 29 2015
  */
 class Language
 {
@@ -27,7 +29,7 @@ class Language
     public function load($name, $code = LANGUAGE_CODE)
     {
         // lang file
-        $file = "app/language/$code/$name.php";
+        $file = "app/Language/$code/$name.php";
 
         // check if is readable
         if (is_readable($file)) {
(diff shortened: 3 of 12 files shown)


andre2:
Why you remove author credits from all files?

jimgwhit:
I feel this is too many changes at once, one file at at time, with no breaking code gives Dave more time to review and test. Look at issue  #174 where volter9 and myself made sure the router worked in all versions.  I feel one file at a time is great plenty here. There are those of us who have actual production databases in place, Dave included.  Any way to close this and do one pull request at a time? You need to leave any current credits.

jimgwhit:
@volter9 is there an email address I can send you a private question? I need some help.

volter9:
@jimgwhit, sure:
```php
$email = $my_github_nickname . 25 . '@gmail.com';

echo $email;
```

volter9:
By the way, @ennerperez didn't removed the credits, he just added his name to any changes he made.

@ennerperez, you don't need to add your credits to all changes you made. Git keeps the information about changes you've made. Everyone who's browsing this git repository will know that those changes you've made.

jimgwhit:
Dave if you implement these are you going to also do, Adjusted invokeObject method in router for PHP7
#175, It will have the framework work on the latest php 7.

jimgwhit:
@volter9 said
~~~


lol, you should change this path in composer.json, not in the file itself :)

{
    /* ... */
    "autoload": {
        "psr-4": {"": "app/"},
        "files": ["app/files/password.php"] /* <- Here */
    },
    /* ... */
}


~~~
I don't think it's an external password library anymore, when ver 2.2 first came out, it was, but then it changed to the internal password file.

volter9:
@jimgwhit even though you're right, It's really _useless_ to modify auto-generated file. What if Dave update the composer files via `composer update` or `composer dump-autoload` and commit it to the repository? The files would be *overwritten*.

daveismyname:
First of all thanks for taking the time to contribute to this framework it's appreciated, as @jimgwhit has mentioned it would be better to do multiple pull requests for each change then they can be accepted/rejected on a per topic basis.

I'm going to reject this PR for the following reasons:

* Uppercase folders in app folder - only folders with classes in should be uppercase.
* this framework uses PRS 1,2 & 4 any code changes should meet the PSR 1 & 2 standards your PR does not, it needs to be modified to comply that way everyone is writing the same style.
* I don't like the socket connection been made on every page load, what's the benefit of setting APPSTATE?
* Never change anything inside the vendor folder as @volter9 has any changes made would be overwritten anytime composer update is ran on composer.json

Adding img to assets might be useful but it needs an alt tag in addition to the file path, it may make more sense to pass an array of options for the image to use so the method can use id, classes etc

The addition db class looks ok but needs to be written to support coding standards for instance rewritten as PSR:

````
try {
    if ($type == 'mssql' || $type == 'sqlsrv') {
        $instance = new Database("sqlsrv:Server=$host;Database=$name", $user, $pass);
    } else {
        $instance = new Database("$type:host=$host;dbname=$name;charset=utf8", $user, $pass);
    }
    $instance->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
    // Setting Database into $instances to avoid duplication
    self::$instances[$id] = $instance;
    return $instance;
} catch (PDOException $e) {
    //in the event of an error record the error to ErrorLog.html
    Logger::newMessage($e);
    Logger::customErrorMsg();
}
````

Placing the routes in another file is an interesting idea that I'm not opposed to I think this should be discussed further to see what other think. If it were to be in a separate file then placing it in the files folder would make more sense then the root.

ennerperez:
Sorry about my mistakes, I just make a pull with multiple suggestions in my particular use; I'll work more organized next time, thanks so much for all comments. I appreciate all explanations. @daveismyname I'll put in suggestions first. Nice to read you.

PR #178, opened 2015-08-29
Add displaySuccess and displayError method

pacozaa:
1. displayError just duplicate the display method for not deprecated the old one
2. displaySuccess is for display a success massage using class= 'alert alert-success'
3. if there is something to show after error it maybe better to have something to show after sucess!

### DIFF
--- a/app/Core/Error.php
+++ b/app/Core/Error.php
@@ -64,4 +64,30 @@ class Error extends Controller
             }
         }
     }
+    public static function displayError($error, $class = 'alert alert-danger')
+    {
+        if (is_array($error)) {
+            foreach ($error as $error) {
+                $row.= "<div class='$class'>$error</div>";
+            }
+            return $row;
+        } else {
+            if (isset($error)) {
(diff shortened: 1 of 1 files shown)


daveismyname:
the current display method can be used for both errors and success message and change the class used by passing the class to be used as a second param, what you've proposed it duplicating functionality.

For displaying a dedicated success message it would be better to use a session that way you can display the success message one time then it clears. That way you don't see the same message again unless the event happens again.

In the sessions helper class is a method called message:

````
/**
     * display message
      * @return string return the message inside div
     */
    public static function message($sessionName = 'success')
    {
        $msg = Session::pull($sessionName);
        if (!empty($msg)) {
            return "<div class='alert alert-success alert-dismissable'>
                    <button type='button' class='close' data-dismiss='alert' aria-hidden='true'>×</button>
                    <h4><i class='fa fa-check'></i> ".$msg."</h4>
                  </div>";
        }
    }
````

pacozaa:
Thank you. sir :)

PR #187, opened 2015-09-11
Add method select row, single and column

felipehertzer:
Added method select only one row, only field, and column from a datebase and added method prepare sql.

Added fetch and fetchColumn.

### DIFF
--- a/app/Helpers/Database.php
+++ b/app/Helpers/Database.php
@@ -77,7 +77,25 @@ class Database extends PDO
     {
         $this->query($sql);
     }
+    /**
+     * A method for preparing before executing the SQL
+     * @param  string $sql       sql query
+     * @param  array  $array     named params
+     * @return array            returns an array of records
+     */
+    private function prepareSQL($sql, $array = array()){
+        $stmt = $this->prepare($sql);
+        foreach ($array as $key => $value) {
+            if (is_int($value)) {
+                $stmt->bindValue("$key", $value, PDO::PARAM_INT);
+            } else {
+                $stmt->bindValue("$key", $value);
+            }
+        }
+        $stmt->execute();
 
+        return $stmt;
+    }
     /**
      * method for selecting records from a database
      * @param  string $sql       sql query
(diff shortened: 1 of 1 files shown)


jimgwhit:
You can get one record already:
```
public function getPet($petid)
    {
        return $this->db->select("SELECT * FROM " . PREFIX . "pets WHERE petid =:petid", 
         array('petid' => $petid));
    }
```
and binding is taking place already: for example:
```
public function insert($table, $data)
    {
        ksort($data);

        $fieldNames = implode(',', array_keys($data));
        $fieldValues = ':' . implode(', :', array_keys($data));

        $stmt = $this->prepare("INSERT INTO $table ($fieldNames) VALUES ($fieldValues)");<---here

        foreach ($data as $key => $value) {
            $stmt-> bindValue(":$key", $value);<-----------------------here------------
        }

        $stmt->execute();
        return $this->lastInsertId();
    }
```
and
```
public function select($sql, $array = array(), $fetchMode = PDO::FETCH_OBJ, $class = '')
    {
        $stmt = $this-> here here here prepare($sql);<--------here-----------------------------
        foreach ($array as $key => $value) {
            if (is_int($value)) {
                $stmt->bindValue("$key", $value, PDO::PARAM_INT);<-------here-----------------
            } else {
                $stmt->bindValue("$key", $value);<------------here------------------------
            }
        }

        $stmt->execute();

        if ($fetchMode === PDO::FETCH_CLASS) {
            return $stmt->fetchAll($fetchMode, $class);
        } else {
            return $stmt->fetchAll($fetchMode);
        }
    }
```
I feel the database helper is fine as is.

geomorillo:
Yes this is already implemented in helper

jimgwhit:
@volter9 how about some input here.

jimgwhit:
Or aa single field
~~~
public function getPet($petid)
    {
        return $this->db->select("SELECT petname FROM " . PREFIX . "pets WHERE petid =:petid", 
         array('petid' => $petid));
    }
~~~

geomorillo:
You can get any field and any record, just implement it n your own model

daveismyname:
I agree with the others the class does everything it needs to do I don't see that adding anything needed, as @geomorillo suggested you can alter your methods within your helper/view.

volter9:
@jimgwhit I think I'm late to share with my thoughts. :cry: 

Although, I think that `raw` should have return statement before it, so people could use returned `PDOStatement` for fetching custom result, etc.

daveismyname:
just to be clear is this what you mean:

````
$data = $this->query($sql);
return $data->fetchAll(PDO::FETCH_OBJ);
````

volter9:
@daveismyname yeah, something like that.

daveismyname:
ah that code results in an error just a simple return does the job.

````
return $this->query($sql);
````

I'll update the class.

volter9:
@daveismyname thanks! :+1:

PR #188, opened 2015-09-14
Added the ability to specify a root directory of engine files

DrTeamHub:
Just change the $smvc variable in index.php file.


geomorillo:
Can you explain why? I don't understand why we need this.

daveismyname:
Yes I'd like your thoughts on this I don't see what the benefit of this is.

volter9:
@daveismyname @geomorillo well, the benefit of this PR is using the absolute path for file `require`ing stuff. On some hostings, there can be a problem with the way how PHP including the files, so absolute path should solve those problems. Also, I think this part of commit is very important:

```php
define('DIR', '/');
```

It's supposed to be committed to the framework, because tweaking the full URL every time isn't helpful, therefore full URL going to bloat the HTML. 

Imagine, I have a website with very long URL. Something like "http://docs.super-simple-framework.com/", and there's menu out of 20 links + 6 links in header. 1014 (26 * 39) extra characters in the output. Using absolute path is better.

geomorillo:
But you can build an url on the html like this echo(DIR."menu1"); where DIR is already defined once in config.php, and  you can require files like this  require(DIR.'test2.php');

volter9:
> require(DIR.'test2.php')

That's. A. Bad. Idea. This line is evaluated to: 'http://domain.com/test2.php' with default setup. `echo DIR . 'menu1';` is Ok, but not requiring stuff.

DrTeamHub:
@geomorillo in my view, store site files (i mean js, css, png etc.) and MVC in same place - it is a very bad idea. For example: https://github.com/DrTeamHub/smvc-drteam
By default in public directory i save only index.php file and some stuff (like js/css), MVC stored in private directory, logs stored in logs directory etc. (plesk-like concept)

daveismyname:
I'm intrigued by the idea, I'm doing a few test on a local setup looking what how this changes the way the classes and modules work if any. 

I do like the idea of using MVC constant for includes.

daveismyname:
@DrTeamHub I'm curious in your version how would you handle having multiple installations on the same domain?

one of my work domains have 20+ different installations each separate from each other as standalone applications that do different things.

DrTeamHub:
@daveismyname 
There are several ways, the simplest is:
private/mvc1/
private/mvc2/
private/mvc3/
publics/test1/index.php (../../private/mvc1)
publics/test2/index.php (../../private/mvc2)
publics/test3/index.php (../../private/mvc3)
The hardest - vhost via proxy eg. nginx proxy_pass or haproxy i think.
Sure there are still ways.

DrTeamHub:
@daveismyname but I prefer to do a: one site - one domain - one mvc

jimgwhit:
@DrTeamHub You do realize you can do it anyway without modifing the  main framework, you're doing it now.  Let's leave the framework as is and any special mods like you want, have your own custom version for that, Thats what I do.

jimgwhit:
Dave, please let's not push this one.

jimgwhit:
I re-read the changes, @DrTeamHub you need to lerrn how to use relative url's. Like:
```
Url::redirect('pet');
```
I don't have to:
```
Url::redirect(DIR . 'pet');
```
Using relative url's is STANDARD PRACTICE in cakephp, laravel, symphony, zend, fatphp SMVC, regular OOP PHP, Java, microsoft.net asp.net, coldfusion / dreamweaver, etc, etc, etc, etc, etc.  It's pull request like this that scares the living daylights out of me.  
Hey by all means use this, but in your own custom version of framework. That's what frameworks are all about, they can be custom tailored.  But please don't ruin my world with a tweak that may only work for you and a special case you may have.

DrTeamHub:
@jimgwhit i just proposed the "root via define" idea :) Good or not, to decide not me. But I think, capability to specify the engine rootdir - it is better than not to have this capability.

jimgwhit:
Are you saying the framework would be setup either way, or would the pull request force your way only???

jimgwhit:
I don't get this line of code:
```
$smvc = '.';
```

jimgwhit:
If pushed, @daveismyname  please have it where a choice is made to which technique is used..

DrTeamHub:
@jimgwhit you're so fast, I don't have time to translate :)
$smvc = '.'; - this line says that the engine is the same place where index.php
if i set $smvc = '../somedir/'; - I mean that the engine is the directory above

DrTeamHub:
@jimgwhit 
mvc (engine) - vendor and app directories i mean

jimgwhit:
Yes, but this can be done on an as needed basis in a custom ver of framework, why change the master framework when it's working now as is?  I think this is one of those things to modify on an as needed basis.  I have seen this asked on laravel forums, how to place the app, etc in a more secure locale, and answers were given.  Your idea is a good one to use in some circumstances, but it should be in a how to topic, not pushed in the framework.

volter9:
> I have seen this asked on laravel forums, how to place the app, etc in a more secure locale, and answers were given

@jimgwhit don't forget that Laravel consists of more than one repository and its skeleton application is just a kick to start working with Laravel core, basically _skeleton application_.

Let's get back to SMVCF. This repository is basically not only skeleton application but also the core, helpers and everything else. I'm not for or against this PR, just saying. :smiley:

jimgwhit:
I have no problems with this technique.  But you could have this part commented out in the framework with instructions on usage.  Then a user could comment out original and un-comment out this and use that technique. It would be easy just to ad the SMVC. where required. That I would totally agree with, let the user of the framework choose here which to use.  
And does this work all all servers, I develop in windows, but linux is the host.

jimgwhit:
Another idea is having an index.md file with instructions on both ways of doing this and why, then the user of the framework could decide, meanwhile leave the original intact. I think a pull request shouldn't change the way things work, like the last router PR.  You could simply copy that router file in, but all still worked.  And the database PR, again other files were not affected.  With this, I'd have to change 40 + files.

jimgwhit:
@DrTeamHub said but I prefer to do a: one site - one domain - one mvc.  
but I prefer  
but I prefer.  
but I prefer.
That's my point, I do not prefer. SO GIVE THE USER THE CHOICE of which technique to use, don't force someone to use something you prefer.  
If both techniques are included with a choice, I have no problems with this.

jimgwhit:
A line like this mixes folder annotation with url annotation:
```
self::loadPlugins(SMVC.'app/Modules/');
```
On a windows test (mine here) machine this translates to:
```
C:\BitNami\wampstack-5.6.2-0\apache2\htdocs\crudv22\app/Modules/
```
Part forward slashes and part back slashes.  Some people still develop on windows.  This does not work. What the  -----?

jimgwhit:
I guess back to the old jream version

jimgwhit:
I went to @DrTeamHub site
```
https://github.com/DrTeamHub/framework
```
Downloaded and tryed on windows. In index where routes are required I added an echo above that line:
```
echo SMVC.'app/Core/Routes.php';
require SMVC.'app/Core/Routes.php';
```
It reads:
```
C:\BitNami\wampstack-5.6.2-0\apache2\htdocs\cwhat\app/Core/Routes.php 
```
But the welcome page came up.  
How in the %$# is the mix of backslashes and forward slashes working?

volter9:
@jimgwhit I infer that PHP fixes those slashes by replacing them on back slashes. There's a link I found about it:
http://stackoverflow.com/questions/14743548/php-on-windows-path-comes-up-with-backward-slash

I don't really know where it's documented on PHP.net

pandory-network:
for replacing back slashes .. btw using linux and windows i used in my project the follow code

```
define('DS', DIRECTORY_SEPARATOR);
```

and i have replaced all / with des Constant DS

daveismyname:
OK I've converted over a live fully functional project into this proposed style now instead of things using the HTTP in DIR it's got the relative path 

````
define('DIR', '/dev/crmcopy/');
````

The same path exists in my .htaccess file.

Everything is works exactly the same as it was before even the modules I have 10 active modules all working as expected I don't see any reason not to accept this. Although the Routes.php should be routes.php as it's not a class.

@jimgwhit I didn't need to change lots of files for this all I changed was 
* index.php (dropped in the SMVC code)
* app/Core/Config.php (updated the DIR)
* app/Core/View.php (adding the SMVC to paths)

That was it to get it working fully.

**Comment by nekomajin** [comment]
I am also in agreement with this proposal. During the development I was messed up often with the paths. So the path specifications are flexible and I think that is what makes a good framework.

**Comment by jimgwhit** [comment]
@DrTeamHub PR changes these.
```
app/Core/Language.php 
app/Core/View.php
app/Helpers/Hooks.php 
index.php 
```
But @daveismyname changes
```
    index.php (dropped in the SMVC code)
    app/Core/Config.php (updated the DIR)
    app/Core/View.php (adding the SMVC to paths)

```
I'm now confused, which of the above is correct?

daveismyname:
I don't use languages so I missed that, the hooks one actually makes sense it's currently:

````
self::loadPlugins('app/Modules/');
````
so it's relative already but adding SMVC to the start makes sense since the app folder can be moved from it's current path with the SMVC constant

I wanted to use it on a live project to see if anything was not working but so far everything works without any problems.

daveismyname:
I've now converted over a CMS built with SMVC, again a very smooth change everything works as normal, I'm going to approve this PR if anything has any concerns please let me know I'm very interested. 

@DrTeamHub can you update the PR and rename Routes.php to routes,php and I will approve this.

DrTeamHub:
@daveismyname done, please check

PR #196, opened 2015-10-02
added key generation method to data helper class

baruchvlz:
I found this to be useful when used with user registration and confirm registration emails.

### DIFF
--- a/app/Helpers/Data.php
+++ b/app/Helpers/Data.php
@@ -85,8 +85,15 @@ class Data
      * key - this will generate a 35 character key
      * @return string
      */
-     public static function create_key()
+     public static function create_key($length = 32)
      {
-       return str_shuffle(md5("!@#$%^&*()_+-=0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHJIKLMNOPQRSTUVWXYZ"));
+        $chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890";
+        $key = "";
+
+        for ($i = 0; $i < $length; $i++) {
+            $key .= $chars{rand(0, strlen($chars) - 1)};
+        }
+
+        return $key;
      }
 }
--- a/app/Helpers/Data.php
+++ b/app/Helpers/Data.php
@@ -80,4 +80,13 @@ class Data
     {
         return ucwords($data);
     }
+
+    /**
+     * key - this will generate a 35 character key
+     * @return string
+     */
+     public static function create_key()
+     {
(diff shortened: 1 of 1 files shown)


jimgwhit:
Does this also work in PHP 7 I ask because as of now the framework is completely PHP 7 compatible and just need to keep it that way?

jimgwhit:
Speaking of which Dave will you be making sure that any future pull request is PHP 7 compatible and backwards compatible to PHP 5.5?

baruchvlz:
I haven't tried it in PHP 7. I honestly haven't read the changes in PHP 7.

jimgwhit:
You said this is for user registration however password is already built into the framework so where and why would you use it just curious.

baruchvlz:
I've used it for database relation keys, confirm email links, and unique keys in rows. I don't like using the a_i id as a way to fetch information from database so I use this to generate a unique key to every user, message, page, comment, post, etc.

geomorillo:
i think we already have a 32 char string generator the method is on Csrf helper
class Csrf {

    public static function makeToken() {
        $csrf_token = md5(uniqid(rand(), TRUE));
        return $csrf_token;
    }

}

baruchvlz:
What version of the framework is that method in? V2.2 makeToken method is different.

```
    public static function makeToken()
    {
        $max_time    = 60 * 60 * 24; // token is valid for 1 day
        $csrf_token = Session::get('csrf_token');
        $stored_time  = Session::get('csrf_token_time');

        if ($max_time + $stored_time <= time() || empty($csrf_token)) {
            Session::set('csrf_token', md5(uniqid(rand(), true)));
            Session::set('csrf_token_time', time());
        }

        return Session::get('csrf_token');
    }
```

geomorillo:
eh you are right maybe this was a previous version sorry.

geomorillo:
I wonder wich one is better 
``` str_shuffle(md5("!@#$%^&*)_+-=0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHJIKLMNOPQRSTUVWXYZ")); ```
or
```  md5(uniqid(rand(), TRUE)); ```

**Comment by baruchvlz** [comment]
Either one works fine. They will both create a 32 char string, so its up to preference.

**Comment by geomorillo** [comment]
also it would be usefull if you could set the string length
```
function randomkey($length = 32) {
        $chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890";
        $key = "";

        for ($i = 0; $i < $length; $i++) {
            $key .= $chars{rand(0, strlen($chars) - 1)};
        }

        return $key;
    }
```

baruchvlz:
+1 
That's a great idea.

Changed the contribution

daveismyname:
thanks really useful!.

PR #198, opened 2015-10-07
Alter variable ADMIN_TEMPLATE not exist in config

felipehertzer:
Change to variable $ custom , as the variable , ADMIN_TEMPLATE is not defined anywhere

### DIFF
--- a/app/Helpers/Url.php
+++ b/app/Helpers/Url.php
@@ -56,7 +56,7 @@ class Url
      */
     public static function relativeTemplatePath($custom = false)
     {
-        if ($custom == false) {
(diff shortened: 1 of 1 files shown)


felipehertzer:
Sorry , for if I ended up making fast and did not see, but the problem is that if you want to define more than one variable , instead of using only constant

baruchvlz:
```
define('ADMIN_TEMPLATE', 'path/to/folder');
define('DEFAULT_TEMPLATE', 'path/to/folder');
```
You add those to your config.php and it should work when calling 
```
Url::relativeTemplatePath(true);
```

PR #199, opened 2015-10-09
List helper

baruchvlz:
Created a list helper that allows for easy creation of ul or ol, li, and a tags.

Use:

```
  use Helpers\Lists;

  Lists::openList(['class'=>'class'], 'ul');
    Lists::openItem();
      Lists::listLink();
      echo "Link";
      Lists::closeLink();
      Lists::openList(['class'=>'submenu'], 'ul');
        Lists::openItem();
          echo "Sub Menu";
        Lists::closeItem();
      Lists::closeList();
    Lists::closeItem();
  Lists::closeList();
```

Renders
```
<ul class="class">
    <li><a href="#">Link</a>
    <ul class="submenu">
        <li>Sub Menu</li>
    </ul>
    </li>
</ul>
```

### DIFF
--- a/app/Helpers/Lists.php
+++ b/app/Helpers/Lists.php
@@ -8,7 +8,7 @@
 namespace Helpers;
 
 /**
- * Create lists and navigations with ease
+ * Create lists with ease
  */
 
 class Lists{
--- a/app/Helpers/Lists.php
+++ b/app/Helpers/Lists.php
@@ -8,7 +8,7 @@
 namespace Helpers;
 
 /**
- * Create lists with ease
+ * Create lists and navigations with ease
  */
 
 class Lists{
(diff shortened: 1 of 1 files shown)


volter9:
@baruchvlz 

> Renders

I think you're wrong, it'll render into something like that: :smiley: 

```html
<ul><li class="item-class"   >item1</li><li    ><a href="http://google.com" class="link-class"  >link1</a></li></ul>
```

I think it's easier to type `<ul>` or `<li>` rather than type long `echo Lists::createList('ul');` or `echo Lists::createitem('item1', ['class'=>'item-class']);` 

P.S.: This file also violates PSR-2, two spaces as indentation, `?>` at the end, random spaces between parentheses in argument declaration, and more than one line breaks at the end of file.

geomorillo:
I agree with volter9 how is this easier than writing ul or li? I lost myself on the second line.

daveismyname:
Thanks but this take more effort to type out then the html tags, This may be best added in as an external add-on once I have that section up and running.

PR #206, opened 2015-11-03
my first attempt to pull requests

gabizz:
Contents:
- Twig helper
- Add port attribute for PDO db connection. (useful when running several Sql instances and/or when deploying desktop apps for customers that are already using port 3306 for other purposes).

### DIFF
--- a/app/Helpers/Database.php
+++ b/app/Helpers/Database.php
@@ -32,22 +32,24 @@ class Database extends PDO
     {
         // Determining if exists or it's not empty, then use default group defined in config
         $group = !$group ? array (
-            'type' => DB_TYPE,
-            'host' => DB_HOST,
-            'name' => DB_NAME,
-            'user' => DB_USER,
-            'pass' => DB_PASS
+			'type' => DB_TYPE,
+			'host' => DB_HOST,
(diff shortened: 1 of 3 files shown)


daveismyname:
thanks for this, as twig requires dependancies (composer to install the library) I cannot approve this PR but this would be a great fit for the new Add-ons section coming this week, it's nearly ready just putting a few final tweaks in places.

The change to the database class would also need a new constant adding to the config class for the port.

gabizz:
Yes Dave, I understand. Haven't expected to merge them, as I told you I'm just figuring out how git works. At least I'll be prepared for future when your modules (addons) will be ready.

And yes, the database class needs the global PORT defined in config, forgot about that, sorry :)

Thank you a lot. I may say I owe you at least half of what I am now as a wanna-be developer. :)

daveismyname:
We teach each other that's how the development community works I learnt far more from the community then I did from tutors at university.

I like to give back where I can, I'll let you know when the new sections is ready in the meantime create a repo on github with the helper, create a readme.MD file with instructions on how to install it and a usage example.

The add-ons section will be pulling in from repo's so all that's needed is the url everything else will be automatic.

PR #208, opened 2015-11-04
Extra route patterns for hexadecimal and uuid v4.

oodgaard:
Thought these might be useful. :all is really not that useful in RESTful pattern matching, and digits (:num) is not the only use case.

If you are using mongo you may want to do:
Router::get('users/(:hex)', 'Controllers\user@getOne');
Router::get('users/(:hex)/reset-password', 'Controllers\user@resetPassword');

If you are using uuid v4 for something:
Router::get('users/(:uuidV4)', 'Controllers\user@getOne');
Router::get('users/(:uuidV4)/reset-password', 'Controllers\user@resetPassword');

### DIFF
--- a/app/Core/Router.php
+++ b/app/Core/Router.php
@@ -63,7 +63,9 @@ class Router
     public static $patterns = array(
         ':any' => '[^/]+',
         ':num' => '-?[0-9]+',
-        ':all' => '.*'
+        ':all' => '.*',
+        ':hex' => '[[:xdigit:]]+',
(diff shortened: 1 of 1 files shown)


daveismyname:
Hi thanks for this how is hex different to num? also what is uuidV4 I've not come across that before.

oodgaard:
Hexadecimal looks like this: 90c46863634ec7040851a324638ca35185c92451
uuid version 4 looks like this: 36fe52a2-77ff-11e5-93f7-a0369f0cea88

daveismyname:
Gotcha thanks for the clarification.

PR #213, opened 2015-11-11
Fix small issues in Pagination

g5pw:
These two commits fix some small issues with formatting/structure to fully adthere to bootstrap's way of doing things.

### DIFF
--- a/app/Helpers/Paginator.php
+++ b/app/Helpers/Paginator.php
@@ -209,7 +209,7 @@ class Paginator
                 if ($this->page < 1 + ($adjacents * 2)) {
                     for ($counter = 1; $counter < 4 + ($adjacents * 2); $counter++) {
                         if ($counter == $this->page) {
-                            $pagination.= "<li class='active'><span class='sr-only'>$counter</span></li>";
+                            $pagination.= "<li class='active'><span>$counter <span class='sr-only'>(current)</span></span></li>";
                         } else {
                             $pagination.= "<li><a href='" . $path . "$this->instance=$counter" . "$ext'>$counter</a></li>";
                         }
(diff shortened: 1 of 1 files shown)


jimgwhit:
Where is the sr-only coming from?

jimgwhit:
Where is the sr-only coming from? How does it work?

g5pw:
The sr-only is a standard bootstrap class, it's used multiple times in the Pagination.php class.
Its contents are hidden to normal users, but shown when a page is read by a screen reader.

I added this patch only because I saw it was used, but in some places it was used incorrectly, and this patch rectifies that.

g5pw:
See [here](http://stackoverflow.com/questions/19758598/what-is-sr-only-in-bootstrap-3) for more information on the sr-only class

jimgwhit:
What is a screen reader? I don't use bootstrap.

daveismyname:
Thanks for this @g5pw @jimgwhit screen readers are software programs that allow blind or visually impaired users to read the text that is displayed on the computer screen with a speech synthesizer or braille display. A screen reader is the interface between the computer's operating system, its applications, and the user.

it's good to support accessibility to make things easier for people with disabilities.

jimgwhit:
Thanks, I learned something new.

geomorillo:
I dont use bootstrap i use semanticui does this affect my css?

daveismyname:
no it's a class being used if you don't have any css for the class it simply won't do anything.

the css for it is this:

````
.sr-only {
    position: absolute;
    width: 1px;
    height: 1px;
    padding: 0;
    margin: -1px;
    overflow: hidden;
    clip: rect(0,0,0,0);
    border: 0;
}
````

More details can be see at http://stackoverflow.com/questions/19758598/what-is-sr-only-in-bootstrap-3?answertab=votes#tab-top

g5pw:
Thanks for the merge and good work on the project! :smile:

geomorillo:
thansk

PR #221, opened 2015-11-26
Changed and added examples of form

BryanYeh:
Added button to examples on the main index page. (you do not want to look at the raw code in welcome/form.php, it's messy)


geomorillo:
Maybe this shouldn't be on the welcome view

BryanYeh:
Well it's just a preview, can always be changed or removed

daveismyname:
I prefer my version of this, been able to set default options in the class is handy without having to set them in the call.

Maybe create this as an alternative forms helper under a new name, that way you can remove me altogether and have the full credit since it's your work afterall.

PR #224, opened 2015-11-30
Improve hookhelper

tomvlk:
Instead of using die() when the $where is unknown it is now throwing an exception. Should be better, and less error prone.
And also removed unneeded else blocks after returning.

### DIFF
--- a/app/Helpers/Hooks.php
+++ b/app/Helpers/Hooks.php
@@ -124,12 +124,11 @@ class Hooks
     {
         if (!isset(self::$hooks[$where])) {
             return false;
-        } else {
-            $theseHooks = explode('|', self::$hooks[$where]);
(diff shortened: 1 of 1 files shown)


daveismyname:
I disagree on this, in the even you use a hook that does not exist it will fail silently. I think it's better to show that hook being attempted does not exist.

I use this within a CMS application the times that's saved me when attempted to use or misspelling a hook name.

tomvlk:
Maybe it's better to throw exceptions instead of returning false then.. But ye, you are right on that. 
Or maybe only die() when we are in production environment?

daveismyname:
An exception would be better I think you're right, if you want to change this is an exception I'll accept that.

tomvlk:
Done :smiley:

PR #225, opened 2015-11-30
added get function to Request and check on SimpleCurl

rtorralba:
I think can be useful that Request has a get method too, maybe there should be a xss filter. At SimpleCurl i added a $params check, because if $params is null or an empty array the $params concatenation is not necessary and can to produce errors.

### DIFF
--- a/app/Helpers/Request.php
+++ b/app/Helpers/Request.php
@@ -27,6 +27,19 @@ class Request
     }
 
     /**
+     * Safer and better access to $_GET.
+     *
+     * @param  string   $key
+     * @static static method
+     *
+     * @return mixed
+     */
+    public static function get($key)
+    {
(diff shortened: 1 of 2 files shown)


jimgwhit:
There's already a get request, it's
```
/**
     * Safer and better access to $_GET.
     *
     * @param  string   $key
     * @static static method
     *
     * @return mixed
     */
    public static function query($key)
    {
        return array_key_exists($key, $_GET)? $_GET[$key]: null;
    }
```
So you are adding something that's there already.

daveismyname:
@jimgwhit is right it's there but under query but as the others have the action as their names it makes sense to have a get too.

I don't want to break this for anyone using query so I'll accept this, no harm in this with it being tiny.

PR #241, opened 2015-12-14
Reworking database helper into better organized database connectors

tomvlk:
To use the database from now on:

```php
$connection = EngineFactory::getEngine();
```
This will return an instance based on your Config.php. 

This structure will allow multiple so called drivers to operate all the same. 
Will also mark the \Helpers\Database as deprecated as it doesn't really belong there anymore, feel free to remove it later.


LuckyCyborg:
@tomvlk Just to note that we started, with Dave's knowledge, the encapsulation of the System components into namespace **\Smvc**. The System refactoring Pull Request incoming in short time, max. tomorrow.

Then, for example, your **\Core\Database** namespace will become **\Smvc\Core\Database**. Be prepared, to not appear conflicts... :smirk: 

All the best!

tomvlk:
Yep, I already read. I will also add a pull request today for a new service implementation. You will love it! :+1: 
When there are conflicts, i'm open to solve it manually.

LuckyCyborg:
You make me curious! :smile:

daveismyname:
how will database calls be made from now on?

tomvlk:
You can use this from now:

```php
$db = EngineFactory::getEngine(); // Will get default driver, from DB_TYPE and all other config.php constants
$results = $db->select("SELECT * FROM " . PREFIX . "car;"); // Handle it just like the \Helpers\Database
var_dump($results); // Result just the same way.
```

Or use the Entity (what was my Service Layer before) for full Object Oriented handling:
** Will be added in other pull request **

daveismyname:
sounds good, what do you think to altering the select to leave out the select so you can do:

````
$db->select("column FROM ".PREFIX."car");
````

tomvlk:
That would also be cool. And that will append SELECT * to the query right?
I will update this pull soon with some more, so please hang on :tractor:

daveismyname:
Yes it's bothered me for ages but didn't want to break what people use but since we are making changes might as well do everything.

kowston:
Yeah go for it, the double select has always bothered me.

tomvlk:
Done :+1:

daveismyname:
thanks!

daveismyname:
Any chance you can provide a bit of an overview of what this brings in?

tomvlk:
Will do a next pull request with more functionality (like my Service thing but reworked) and this will highly works on this.

But in general this will provide in resolving the issue when using other databases like PostgreSQL or smilary with the framework. The interface on how to use the Engine is always the same for every *Engine there will be. So it will work (almost) the same for MySQL as for PostgreSQL (I'm pretty sure this needs testing and tweaks when doing any).

tomvlk:
Also see https://github.com/simple-mvc-framework/framework/pull/243

PR #243, opened 2015-12-14
Database Service Layer, working on the new Database Engines.

tomvlk:
This will make the \Core\Service be as a generic service. And add a DatabaseService with already there the MySQLService including for using Entities and Services to act like a small ORM. 

To get started:
- Create an entity in App\Models\Entities that extends the \Core\Database\Entity.
- Add all public $vars to match your database model (look at the example of car).
- Create an service in App\Services\Database that extends any of the driver specific Services (now only \Core\Database\Service\MySQLService.
- Make sure you will implement the __construct() with optional a custom engine in the parent constructor.
- Make sure you set the following variables in the __construct(): 
```php
        $this->table = "car"; // Table name, without prefix!
        $this->primaryKeys = array("carid"); // Primary key(s)
        $this->fetchMethod = \PDO::FETCH_CLASS; // Fetch method
        $this->fetchClass = '\App\Models\Entities\Car'; // Full namespace to the attached entity you made earlier
```
- Default you don't have to add more functions to support CRUD operations. But for example you could add a select all function:
```php
    public function getAll()
    {
        return $this->read("* FROM " . PREFIX . "car"); // Make sure you use the prefix here!
    }
```

Now you are done! You created your ORM Service and Entity class. You can use it like this:

```php
        // Use with the Car service:
        $service = new Car(); // \App\Services\Database\Car is the full namespace path
        $result_service = $service->getAll();
        var_dump($result_service);
```
And it will return an array with Entity instances.
You can now (for example) delete a single entity or more with:
```php
        $service = new Car(); // \App\Services\Database\Car is the full namespace path
        $result_service = $service->getAll();
        var_dump($result_service);
        $service->delete($result_service[0]);

        // Test by getting it again and print again.
        $result_service = $service->getAll();
        var_dump($result_service);
```

Also added a sample, you can import database with the attached file.

[cardemo_database.sql.txt](https://github.com/simple-mvc-framework/framework/files/61347/cardemo_database.sql.txt)


LuckyCyborg:
@tomvlk The pull request for System's new namespace, aka **\Smvc**, is in pulls line, please align your namespaces too... :wink:

tomvlk:
Yep, I will apply it as patch and solve the conflicts :+1:

tomvlk:
Better to first approve the namespace restructure then this pull.

LuckyCyborg:
What happened with this pull-request? Problems after System namespace change?

tomvlk:
Alreay had conflicts before that.

tomvlk:
Ok, solved. Wasn't merged for some reason, I really thought I already did. Anyway, it's fixed.

LuckyCyborg:
Something is wrong with your **/vendor** directory. It should not be now in pull requests and repository.

tomvlk:
Also fixed.

LuckyCyborg:
I seen! :+1:

LuckyCyborg:
I read you pull-request, as how is now. Yep, That's what I call **Database Service**...

PR #247, opened 2015-12-14
Moving sample from index page to the demo page with route /database.

tomvlk:
You have to import the testdata from the sql export:

[cardemo_database.sql.txt](https://github.com/simple-mvc-framework/framework/files/61510/cardemo_database.sql.txt)


LuckyCyborg:
That's better! :+1:

PR #248, opened 2015-12-14
Refactoring of SMVC Configuration.

LuckyCyborg:
Initial commit for SMVC Configuration refactoring.

I proposed the very simple class for configuration, mentioned also in my first post:

```php
namespace Smvc\Core;

class Config {

    protected static $settings = array();

    public static function get($key) {
        return isset(self::$settings[$key]) ? self::$settings[$key] : null;
    }

    public static function set($key, $value) {
        self::$settings[$key] = $value;
    }

}
```

It's usage is very simple, and it is very versatile. For example, if you want to do a active Modules list, you can do:

```php
Config::set('modules', array(
    'System',
    'Cron',
    'User',
    'Menu',
    'Page',
    'Template',
    'MediaManager',
    'Contact',
    'WebChat',
    'CDN',
));
```

We can store in it even arrays, in fact, we can store everything that can be stored in a array, and the built-in behavior is global read access.

Of course, this is not a replacement of the constants (aka defines), but a companion.

We can have the following configuration files:

```
app/config/bootstrap.php
app/config/config.php
app/config/constants.php
app/config/routes.php
```

and in **public/index.php**, instead of old standard Config instantiate, just:

```php
require SMVC.'config/bootstrap.php';
require SMVC.'config/routes.php';
```

The **SMVC/config/bootstrap.php** can contains some global useful functions, to include the **config.php** and **constants.php** and to do the initialization stage. For example, can contains:

```php
use Smvc\Helpers\Session;

/**
 * Turn on output buffering.
 */
ob_start();

require dirname(__FILE__).'/constants.php';
require dirname(__FILE__).'/config.php';

/**
 * Turn on custom error handling.
 */

set_exception_handler('Smvc\Core\Logger::ExceptionHandler');
set_error_handler('Smvc\Core\Logger::ErrorHandler');

/**
 * Set timezone.
 */
date_default_timezone_set('Europe/Rome');

/**
 * Start sessions.
 */
Session::init();
```

The **constants.php** and **config.php** files content is easy to suspect.

That's all! We have a very simple **SMVC's Configuration**, which can be extended easily, dedicated per (every) application(s)...


tomvlk:
Love the final fix :rocket: :smile: 
Will update the mailer, database etc asap after merge.

LuckyCyborg:
@tomvlk You know what is very nice? For example, my little **\Smvc\Core\Config** class, with two static methods, permit to have, for example, multiple database configs. I.e.

```php
Config::set('database', array(
    'default' => array(
        'username' => 'tomsdb',
        'password' => 'password',
        ...
    ),
    'sqlite' => array(
        'filepath' => 'shinypath'
    ),
    ...
));
```

tomvlk:
You didn't add the configs to the gitignore? And make sure you make it config.example or something

tomvlk:
And yes, I will look into the database config stuff, maybe make it like this:
```php
array(
'default' => array(
  'engine' => 'mysql',
  'blabla auth etc etc' => 'okok'
),
'other' => array(
   ... etc...
)
);
```
So you could give the name of the database connection to the engine factory.

```

LuckyCyborg:
@tomvlk I didn't added any file to .gitignore or something... Let's see what path go Dave, in principle can be two files as "candidates" for examples: **/app/Config/config.php** and **app/Config/constants.php**,  one dynamic, another one static, but lets see his decisions...

tomvlk:
Yep. Thanks for the nice work you did everywhere :+1:

PR #249, opened 2015-12-14
Moving database from \Smvc\Core\Database to \Smvc\Database

tomvlk:
Moving database from \Smvc\Core\Database to \Smvc\Database


LuckyCyborg:
@tomvlk Look there: #248

There is something which I bet you will like, as **Mailer** configuration...

PR #252, opened 2015-12-15
Optimize and speed-up the ClassicRouter, other Routing small fixes.

LuckyCyborg:
Some Routing small fixes and **ClassicRouter** optimization (for speed).


daveismyname:
Looks good should the new router by on by default instead of the ClassicRouter?

LuckyCyborg:
@daveismyname My bad! I will fix right now!

PR #254, opened 2015-12-15
The prefixing 'SELECT ' in the query (for mysql engine) can still be in the query.

tomvlk:
Now you can do it with or without:

```php
$result_plain = $engine->select('SELECT * FROM '.PREFIX.'car');
$result_plain = $engine->select('* FROM '.PREFIX.'car');
```

Remeber, we need to move the configs and add .dist.php after it, and add it to the gitignore when stable. So we could update the git without having to clear our personal configs each time.

### DIFF
--- a/app/Controllers/Demo.php
+++ b/app/Controllers/Demo.php
@@ -52,6 +52,10 @@ class Demo extends Controller
 
         // Use it without the Services:
         $engine = EngineFactory::getEngine();
+        $result_plain = $engine->select('SELECT * FROM '.PREFIX.'car');
+        var_dump($result_plain);
+
+        // Using the select and prefix the SELECT in the sql is optional for the MySQL Engine!
         $result_plain = $engine->select('* FROM '.PREFIX.'car');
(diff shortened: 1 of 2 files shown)


daveismyname:
ah you beat me to it! :+1:

tomvlk:
Haha, sorry

nhymxu:
if using ->select()
why don't use ->select('column', 'table', 'where'); ???

Clean structure

tomvlk:
That would be more a query builder, which this isn't really.

daveismyname:
plus that wouldn't work for joins without more work. Going down that route it would be easier to just use eloquent I think.

nhymxu:
I don't thing **->select("* FROM table")** more clearer than **->query("SELECT * FROM table");**

If you want seperate **->query** and **->select** **->insert**. Just do more.
See WordPress wpdb. support **->select** **->insert** method, and classic **->get_results("SELECT * FROM ... )**

PR #259, opened 2015-12-15
Splitting the \Core\Error controller in a real App Controller and an …

LuckyCyborg:
…Error Messages generator; also several another fixes.

The main feature of this pull-request is the splitting of **\Smvc\Core\Error** controller in a pure App Controller, located in **\App\Controllers\Error** and remaining class containing only the static method which generate Error Messages. A rationale for this move:

- The Error Controller's View is located in App's Views folder already, and moving the Controller part also in App, will simplify further developments of Router/Controllers/Views flow.
- Isolation of SMVC's System from App Tasks. No one should touch the System folder beyond the SVMC developers, while having the Error Controller located in App, the end-user can customize it as he like, eventually extending for custom Error Messages. 

In other hand, following the style of another directories from **App** directory, and for a consistent design, I renamed **app/views** uppercase-first, then **app/Views**, also, I create a match between the name of associated Controllers and the directories who correspond in **Views**. The rationale being that this way is much easy to manage the structure when there are a huge numbers of Controllers, whitout wondering which one correspond with one directory.


daveismyname:
OK that makes sense I may be asking a lot of questions about these recent changes towards the end of the week when I start going through the updates so I can test and use the beta.

LuckyCyborg:
Sure! There I'm to respond to your questions! :smile:

Also, I want to assure you that every change proposed to SMVC is well debated and considered before arriving there. As you remember, principally, I'm the designed Speaker, but the decisions I don't make alone. ;)

PR #263, opened 2015-12-16
Introducing the Modules Management and their bootstrapping

LuckyCyborg:
This push-request introduce the Modules Management to SMVC 3.0

Right now, the Modules Management execute the bootstrap of active Modules, on the Framework boot stage, permitting to have and use **per Module** configurations. 

The Config entry for the active Modules, from **app/Config/config.php** is in an array of strings, containing the Modules name.

Typically, a file located in Module's **Config** directory, called **bootstrap.php** is executed if exists. For example, for a Module called **Blog**, its bootstrap file is located in:

```
/app/Modules/Blog/Config/bootstrap.php
``` 

The Module's bootstrap files should call the companion files, usually **config.php** and **routes.php**. When new constants, specifically for this Module are defined, they go on **constants.php**

An example content for a typical bootstrap file is:

```php
//require dirname(__FILE__) .'/constants.php';
require dirname(__FILE__) .'/config.php';
require dirname(__FILE__) .'/routes.php';
```

The second stage **config** files content are as usual, similar with those from App Configuration.

### DIFF
--- a/app/Config/bootstrap.php
+++ b/app/Config/bootstrap.php
@@ -7,8 +7,9 @@
  * @date December 15th, 2015
  */
 
-use Smvc\Helpers\Session;
+use Smvc\Core\Config;
 use Smvc\Core\Router;
+use Smvc\Helpers\Session;
 use Smvc\Modules\Manager as Modules;
 
 /**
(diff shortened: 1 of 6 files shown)


tomvlk:
Really nice addition!

LuckyCyborg:
@tomvlk  Thanks! :smile:

PR #277, opened 2015-12-17
Adding SQLite Engine (pdo) and improving engine structure.

tomvlk:
All tested, but please test too :+1: 
Also added sample config, and adding constant for using with the factory.

### DIFF
--- a/app/Config/config.php
+++ b/app/Config/config.php
@@ -65,7 +65,16 @@ Config::set('database', array(
             'user' => 'root',
             'password' => 'password'
         )
+    ),
+    /** Extra connections can be added here, some examples: */
+    /*
+    'sqlite' => array(
+        'engine' => 'sqlite',
+        'config' => array(
+            'file' => APPPATH . 'database.sqlite'
+        )
     )
+    */
 ));
 
 /**
--- a/system/Database/Engine/SQLiteEngine.php
+++ b/system/Database/Engine/SQLiteEngine.php
@@ -1,0 +1,197 @@
+<?php
+
+
+namespace Smvc\Database\Engine;
+
+
+class SQLiteEngine extends \PDO implements Engine
+{
+
+    private $config;
+
+    /**
+     * SQLiteEngine constructor.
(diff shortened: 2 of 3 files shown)


LuckyCyborg:
@tomvlk I propose to move the common methods from:

 **\Smvc\Database\Engine\MySQLEngine** and
 **\Smvc\Database\Engine\SQLiteEngine**

In a **Trait** called, for example:

**\Smvc\Database\Engine\CommonLayerTrait**

then, to use it in both engines. 

This way, you can avoid the code duplication and simplify the maintenance... :smirk:

tomvlk:
@LuckyCyborg 
The trait idea, great, but for some operations sqlite is a bit different than mysql. And to be fair I have to experiment with it first. I could do some refactor and make the engines better and generic as possible. So less code is required.
Maybe have few interfaces or something, at least It can be better indeed. But requires more classes/traits/interfaces etc, harder to understand. I think the new engines should be very well documented on release.

Btw, the storage of the database is just a full path. You can see in the config that you are required to put the full path. But default it's using the app path.

(Please wait with merging, I will improve it futher).

LuckyCyborg:
@tomvlk I added for you, the usual **file storage path**, for non-temporary one, in my current pull-request.

It is: **/storage**, as usual

Let's do not suggest to end-users to do naughty things... :smirk:

tomvlk:
@LuckyCyborg Did you add another folder? Use that for strorage?

LuckyCyborg:
@tomvlk In the **/storage** are preserved files who are **not temporary**. For example,your SQLite database file.

tomvlk:
Ahh ok, but its still in the unaccepted pull, I see. Better that indeed. Force to storage is best option.

tomvlk:
@LuckyCyborg I think I will make an interface that every engine will implement, and will have the basic things like read($sql) or something. Will look into it later.

LuckyCyborg:
@tomvlk OKay. :smirk:

PR #281, opened 2015-12-19
Database Engine and Service refactored.

tomvlk:
# Changes. Merge Ready.
**The following actions have been taken to improve the overall structure of the new Database Engines.**
- [x] Adding a generic interface to have generic and non engine dependend functions for executing queries.
- [x] Making the services work with the new engines. Adding support for multiple inserts.
- [x] Adding a demo for all the newest changes.
- [x] Adding more configurable stuff for engines
- [x] Providing better PHPDoc.
- [x] Providing legacy calls on mysql engine.
- [x] Adding raw prepare function on PDO engines.

## Function changes
Basically, only the functions are a bit different.

## Changes on replaced functions
There are only some adjustments on the above replaced functions, listed bellow:

#### select
- Will be using the default fetch method that you gave in the config when not giving in the parameter.
- When no fetch method is given in the config it will use the FETCH_OBJ.
- Will return better error status, return false on failure earlier than before. Also on empty response it will return false.

#### insert
- Can handle **Multiple Inserts**. You can do this by giving a multidimensional array in the $data parameter.
- Can handle transactions, and will rollback if 1 error is discovererd. Default it is not being used!
- Will return multiple inserted id's when giving multiple inserts.

#### update
- On the mysql engine it will you can use the limit parameter.

#### delete
- On the mysql engine it will you can use the limit parameter.

#### NEW: raw / rawQuery
Can execute simple queries, with no extra binding values. Will return an array or false.
For example you can use this to truncate the table.
executeSimpleQuery("TRUNCATE table;");

#### NEW: rawPrepare
Will return a prepared statement, for all PDO drivers this will be a PDOStatement object.

## MySQL Service Changes
- create: Can handle multiple inserts too.

## The old helper:
The old helper is now a wrapper around the MySQL engine.

## Demo's
See the file app/Controllers/Demo.php for some examples on how to use.
We need to make better examples later.


LuckyCyborg:
@tomvlk Well, while I have no special interest about the internal names on Database Engine, permit me to disagree with the name changes on **API** interface.

First of all, there should be a backward compatibility with the old API, to not break completely the things in porting to the new version, second, the names like **select, insert, update, delete** are well known&used on different DBAL engines. 

Please return back to the old ones, because I consider that changing them in this way not being the most luckiest idea. Or you expect us, for example, to rewrite hundreds of Models just because?

All the best!

tomvlk:
@LuckyCyborg I could create the select, insert, update and delete as a alias for the new functions.
Hold on .. will implement that. Good point btw.

LuckyCyborg:
@tomvlk And the old \Nova\Helpers\Database instead to being nuked, as a transition solution, it can be rewritten as a **Wrapper** on the new API, preserving the maximum compatibility of the old code.

tomvlk:
@LuckyCyborg Updated! :+1:

LuckyCyborg:
@tomvlk Good to see them, but just nitpicking that **execute*** shall be considered internal names (i.e. protected) and the public one will be as suggested. And there is no problem to imagine, for example, a backward compatible **update()** method, bo offer the new extensions too... :smirk:

LuckyCyborg:
@tomvlk And you need something like **rawQuery** and **rawPrepare**, very useful for big databases.

Think about billions records living in a table. We will map all in memory on **select**?

tomvlk:
@LuckyCyborg Let's improve it a bit more, 

please wait before merge...

tomvlk:
I think the rawStatement is already enough, as the MySQLEngine extends PDO itself you could also execute all pdo stuff on it.

LuckyCyborg:
@tomvlk Nope, I can give you dozens of examples about using **rawQuery** and **rawPrepare**

rawQuery when you do diectly, well... simple queries, for examples. It is faster.
rawPrepare when you do prepared statements, and joins, and other thingies.

And those are another well used names. :smirk:

tomvlk:
You can use the executeSimpleQuery for it. I improved it so you can use it without having a fetch result going back, this will only execute and give boolean if succeeded. (like the $pdo->exec();)

LuckyCyborg:
@another thing: implement a query counter.

It is a simple thing, but very useful. You have a internal private variable, default to 0 in class construction then, every query, it increments. Finally, you have a method to read the counter.

Secondly, implement the option of connection compression, with 
```
PDO::MYSQL_ATTR_COMPRESS => TRUE
```
When the Database server is different of that of the Web-server, the response speed-up is spectacular.

LuckyCyborg:
If you accept to wait until Monday with the merging the pull, I will consult our Database's specialist on your actually Design.

tomvlk:
@LuckyCyborg That would be great. 
Maybe also great to head other people opinions about the api. 
Off coarse, smaller function names are better, but it could also conflict with any extending function. 
Anyway. Thanks for your feedback :D.

LuckyCyborg:
@tomvlk also **charset** should be optionally added, when is defined. I know Database Servers who do naughty things when it is specified.

nhymxu:
I hate long function name. $engine->executeQuery look clean function name but not friendly and tired when typing

tomvlk:
@nhymxu I must agree sometimes, but I'm used to have functions that are really long. Also is better but not friendly for typing.
Maybe I could indeed do a public api with shorter, or the same like it was by simply renaming the trait from legacy to public.

LuckyCyborg:
@tomvlk Meantime, implement, please, the **Query Counter**. It is the base of debugging, later we will improve the DBAL debugging, but this counter I consider a must to have.

LuckyCyborg:
@tomvlk Looking at your **Nova\Database\Engine\Engine** and **Nova\Database\Service\Service** , what I see are typically cases of **Contracts** or **Promises**.

I suggest you moving them in **Nova\Database\Contracts** namespace. Will be more clean this way.

LuckyCyborg:
@tomvlk In other Hand, why we should have **Nova\Database\Engine\MySQLEngine** while is more facile to have: **Nova\Database\Engine\MySQL** ?

LuckyCyborg:
@tomvlk Also **Nova\Database\DatabaseService** can be simply: **Nova\Database\Service**

Please, think in namespaces, as in a class description. Further, we can use as final end-user of the Service, something like:

```php
use **Nova\Database\Service** as DatabaseService;
```

tomvlk:
@LuckyCyborg Ye. I will send more commits to improve tonight.

LuckyCyborg:
@tomvlk Sure, Tom!  :smile: 

I can't wait to see your next development!

Do your best of! :+1:

geomorillo:
Good luck :+1:

tomvlk:
The refactor of namespaces is done, it's much simpler and better.
Using the old style functions.

@LuckyCyborg How do you count it? Also each query inside of a transaction? Also failed queries?

LuckyCyborg:
@tomvlk The Query Counter is a very simple concept, like I said. And spectacular useful for developers, because a significant part of optimization is about minimizing the number of Database Requests (Queries).

In princple you have a:

```
class ShinyDatabase 
{
    private $totalQueries = 0;
```

then for every, but every query possible to be counted, even it failed or not, as:

```
    public function rawQuery($sql)
    {
        $this->totalQueries++;

        return $this->query($sql);
    }
```

see that we need **rawQuery()** ? :smirk: 

Finally, a simple getter:

```
    public function totalQueries()
    {
		return $this->totalQueries;
    }
```

In other hand, looking at your new classes structure is it much much better! :+1:

LuckyCyborg:
@tomvlk Look there to see something cool:

https://github.com/LuckyCyborg/simple-mvc-framework/tree/3.0-events-management

tomvlk:
Aha. Will work on that.
Btw: PHP passes Arrays and Objects by reference (pointers). So you don't need & when saving or getting instances.
http://php.net/manual/en/language.oop5.references.php

LuckyCyborg:
@tomvlk Thinking something...

How about rename the **\Nova\Database\EngineFactory** as **\Nova\Database\Manager** ?

That way we will be consistent with another Managers:

\Nova\Events\Manager
\Nova\Modules\Manager

tomvlk:
@LuckyCyborg Would be cool I think, and a bit better than factory. Well ye it's an factory but Manager is also fine, and a bit more generic.
If you change yours I'll change mine :laughing: .

LuckyCyborg:
@tomvlk \Nova\Core\Controller is a special kind of animal. The Controller is designed to work as unique instance called on Flight.

When the Event is deployed to a Controller, it will change the instance, then, without properly restoring, the \Nova\Core\View will go crazy! :smirk:

tomvlk:
@LuckyCyborg I saw it on the language in core.
Anyway, the query counter is in. But also counting on rawPrepare and raw(Query). If that isn't mostly the case then tell me, ill change it.

LuckyCyborg:
@tomvlk The Counter should count, well ... everything, every Query possible to be counted, in every method.

tomvlk:
@LuckyCyborg Then it's ok right now :D. Only if you are using the PDO functions instead of the engine it's not counting, but ye, thats not recommended of course.

LuckyCyborg:
@tomvlk :smirk:

tomvlk:
@LuckyCyborg Btw, I think that the Helper Assets is not an helper anymore. Am I right? It's still used in the js and css right?
Maybe an idea to implement that into the view.

tomvlk:
@LuckyCyborg And session helper is more something for Net\ namespace.?

LuckyCyborg:
@tomvlk In my opinion, Dave got right the idea with the Assets helper, as in a Engine to pre-process on-fly the Asset Files, and automatically caching them minimized is a plus. :smirk:

LuckyCyborg:
@tomvlk But, yeah! Session helper should go on **\Nova\Net** namespace, as meaning.

When we enter in Final Phase, I will take care about those things. 

Now I have to finish the **Events Manager**, who literally live on **\Nova\Events\Manager** class.

LuckyCyborg:
@tomvlk Hey! Now the DBAL works considerable better! Congratulation! :+1:

tomvlk:
@LuckyCyborg Still wait on feedback until monday, or merge and do a fixup/refactor after monday?

LuckyCyborg:
As i see now, after I read your code, looks good.

In think to do the merge, and Monday I will ask my colleague for critics and suggestions.

tomvlk:
Ok, Then @daveismyname can now pull the trigger ehhhh request I mean. :+1: 
Hit that nice shiny green button :laughing: .

daveismyname:
roger that! :+1:

LuckyCyborg:
@tomvlk BTW, you do not write a **Service** for **SQLite** too?

tomvlk:
@LuckyCyborg Ye I will. But not now.

geomorillo:
Its going SuperNova ....

PR #283, opened 2015-12-19
Events Management for Nova Framework

LuckyCyborg:
The proposed **Events Management** implements a simple solution of **Event** deploying to Closures and Class Methods, even to live Controllers.

When the **Event Listener** is a Controller, its listening **Method** is executed in a fully configured **Views Context**, similar with the **Routing** case, even when the called Controller live in one of **Modules**.

Then, include the ability to collect information and data from different stand-alone Modules, thing very useful, for example, in the case of implementing Page **Blocks**.

Usage is very simple, to the **EventManager**, living as **\Nova\Events\Manager**, we add Listeners using a static way or normal one, i.e.:

```php
use Nova\Events\Manager as EventManager;

EventManager::addListener('welcome', 'App\Controllers\Demos\EventListener@welcome');
```

Like you can see, the syntax is very similar with one used by **Routing**.

Then, to trigger an Event **deploying**, we can do something similar with:

```php
    private function fetchEvent($event, $params)
    {
        $eventManager = EventManager::getInstance();

        // Trigger the Event and capture the result.
        $data = '';

        $eventManager->trigger($event, $params, function($result) use (&$data) {
            $data .= $result;
        });

        return $data;
    }
```
More simple, for a shoot-in-the-dark of the Event:
```php
$eventManager = EventManager::getInstance();

$eventManager->trigger($event, $params);
```
OR using the static one:
```php
EventManager::sendEvent($event, $params);
```

Finally, a **Listening** Class Method or Closure is an usual one, but it should communicate with the *EventManager** returning a Result of its call. Example:

```php
    public function welcome($event)
    {
        $params = $event->params();

        $message = __('Hello, welcome from the welcome controller! <br/>
This content can be changed in <code>{0}</code>', $params['path']);

        return $message;
    }
```
To note that the Listening Method receive as parameter a **\Nova\Core\Event** instance, from where being extracted eventually the Event Name and Parameters.

The implementation come with a complete example, located in **\App\Controllers\Demos**

@daveismyname You can merge the pull-request, please!


LuckyCyborg:
@daveismyname Houston to Dave! Houston to Dave! 

Please, merge this pull-request to end our Development Stage!

PR #287, opened 2015-12-21
3.0 - Complete the Framework components integration

LuckyCyborg:
**WIP - DO NOT MERGE**


daveismyname:
just checking you still not ready to merge correct?

LuckyCyborg:
@daveismyname No, is not ready, thank you! :smile:

Talking about **\Nova\Helpers\Hooks**. 

This Helper functionality is superseded now by **Events Management**, present starting with previous commit, who can do, everything what this helper can do, while offering a much powerful support for. 

Also, now the Controllers have a internal Data Storage, where can be configured diverse variables, and much more important, **\Nova\Core\Controller** and **\Nova\Core\View** works in duet, for properly rendering, the **View** needing a properly initialized **Controller**.

Even the Modules Management do not need this Helper, right now an Module having a complete configuration sub-system, where one can setup per-Module configuration parameters, constants or routes.

Then, as is it, the **\Nova\Helpers\Hooks** not only remained under the Framework development, but also can create subtle conflicts and make *the duet Controller/View* to go crazy.

I'm on opinion about nuking the actual **\Nova\Helpers\Hooks**, as being obsolete, and pass to using the Events instead. 

Eventually, we think about to write a compatibility Class wrapper, as **\Nova\Modules\Hooks**,  who emulate the old Helper functionality on top of Events and its EventManager.

To see how it works the things without Hooks, while using Events to request data from another Classes or even Controllers, look at my current branch:

https://github.com/LuckyCyborg/simple-mvc-framework/tree/3.0-final-stage

And an example page who request its content from another Controller is it there:

http://smvc3.giulianaeassociati.com/demos/events

daveismyname:
can you point to where I can see inject code from one place to be executed in another. I can see this, but unsure what is it's doing.

````
$params = array('path' => $this->filePath);

// Get the Message.
$message = '';

EventManager::sendEvent('welcome', $params, $message);
````

For instance in one of my modules I have hooks like these:

````
Hooks::addHook("modules", "\Modules\Broadband\Controllers\Broadband@addOns");
Hooks::addHook("topnav", "\Modules\Broadband\Controllers\Broadband@nav");
Hooks::addHook("sidebarnav", "\Modules\Broadband\Controllers\Broadband@nav");
Hooks::addHook("routes", "\Modules\Broadband\Controllers\Broadband@routes");
````

How would I do that with Events?

**Comment by LuckyCyborg** [comment]
@daveismyname You spotted right where the Event is handled. So, $params is always an array of parameters who is passed to Listener, via an (simple) **\Nova\Core\Event** instance who store a name (of the Event) and its parameters.

Al listener Closure or Class Method receive this Event as parameter, in execution. Look into **\App\Controllers\Demos\EventListener::welcome** method. It is a typical case of Listener.

A Listener function can return a **Result** back to **EventManager**, who pass it to a **Callback**, for every Listener iteration. Main processing method is **\Nova\Events\Manager::notify**, but there are also diverse helper methods, and right now I work even in a Hooks Compatibility Layer.

One of very useful composite methods is **EventManager::sendEvent** who pass to the Listeners registered for the specified Event a parameters array() and smart collect the Results from Listeners. The third parameters is $result, which can be an array(), case when the Results are inserted in it as values, like $results() = $result, or can be a string, case when the Results are collated in an appending way.

For example, if you want to collect a **string** result from the 'topnav' Event's Listeners, you should have:

```php
$params = array(<shiny things there>);

$result = '';

EventManager::sendEvent('topnav', $params, $result);

echo $result; // Contains the Listeners work.
```

Sure, similar with how you registered the Hooks, you need to register the Listener to Event. Then, in **/app/Config/Config.php** or, for a Modules, **/app/Modules/Blog/Config/config.php** you should have, very similar to Hooks one:

```
EventManager::addListener("topnav", "\Modules\Broadband\Controllers\Broadband@nav");
``` 

And Voila! 

Your Listener Closure or Method will receive the **\Nova\Core\Event** instance **$event** as parameter, when you deployed the Event via **EventManager::sendEvent()** !

In other hand, you do not need to define via **Events** the "modules" and "routes", the best way 
to define them is using the native now Configuration. First one in **/app/Config/Config.php**, second one in, ie **/app/Modules/Blog/Config/routes.php** 

For a more advanced usage of Events handling look to **\App\Core\BaseController**, who deploy an Event in its beforeFlight and collect the data from Listeners, to prepare its internal Data entries.

Attaching Listeners to this Event you can add information to all Controllers, about **CSS** files and **Javascript** ones, in a single **Event** shoot (loop), information who finally arrive on **\Nova\Core\View** Data and are available on Views. 

I choose to create a BaseController instead of integrating this support directly in \Nova\Core\Controller because there is the possibility to have API controllers who not need GUI Data loaded.

These are some examples about how to use Events. If you want more details, just ask; our new Events Management while simple, is powerful, and I believe that you'll love it.

daveismyname:
OK I need to sync down the recent changes and have a got at getting my head around this so I know what I'm talking about better.

btw thank you for all your efforts it's really appreciated.

LuckyCyborg:
@daveismyname The, I propose you to merge this pull-request, it is stable enough to consider being even in a Beta, more it is for tuning. And I will open tomorrow another develop-PR.

That's way you can see latest implementations, BUT be warned! your Hooks are nuked from orbit. :smirk:

daveismyname:
OK I understand.

PR #291, opened 2015-12-22
3.0 fix email

nekomajin:
@LuckyCyborg The only change is the correction of the double @@ characters :smile:

### DIFF
--- a/app/Config/bootstrap.php
+++ b/app/Config/bootstrap.php
@@ -2,7 +2,7 @@
 /**
  * Bootstrap handler - perform the Nova Framework's bootstrap stage.
  *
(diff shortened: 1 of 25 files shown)


LuckyCyborg:
@nekomajin Huh! Was a little anti-spam "protection". :smirk: 

But, if the community wants the change, why not?

nekomajin:
Ok, sorry. This makes sense of course. It's @daveismyname's decision whether he merge it.

LuckyCyborg:
@nekomajin No problem! If we really want us to be a little pedantic in emails, I have nothing against this change... :smile:

PR #295, opened 2015-12-23
Add method to handle PUT and DELETE HTTP method

cahyadsn:
- Add method to access PUT and DELETE parameter and to detect if request is PUT or DELETE request
- improve post() and get() method capabilities to handle access for all POST/GET parameters if no key parameter provided


tomvlk:
May i thank you for adding this! :+1:

PR #296, opened 2015-12-23
style adjustment for message type

cahyadsn:
currently this method only provide alert box with class="alert alert-success alert-dismissable" even the $sessionName is 'info','danger' or 'warning' (base on 4 alert box type on Bootstrap). With little change than this method can show alert box with class="alert alert-info alert-dismissable" , "alert alert-danger alert-dismissable" or "alert alert-warning alert-dismissable" depend on giving $sessionName parameter

### DIFF
--- a/app/Helpers/Session.php
+++ b/app/Helpers/Session.php
@@ -171,7 +171,7 @@ class Session
     {
         $msg = Session::pull($sessionName);
         if (!empty($msg)) {
-            return "<div class='alert alert-success alert-dismissable'>
+            return "<div class='alert alert-{$sessionName} alert-dismissable'>
                     <button type='button' class='close' data-dismiss='alert' aria-hidden='true'>×</button>
                     <h4><i class='fa fa-check'></i> ".$msg."</h4>
                   </div>";


tomvlk:
@cahyadsn Maybe you should also check for the type of alert, if it's set or not. And also add the
```html
role="alert"
```
to the end of the alert div, just as it should be according the docs of bootstrap.

daveismyname:
the problem with this approach you loose the ability to have a custom message in the alert div. The $sessionName needs to be one of the 4 types to apply the color but then it's also used to show inside of the div results in cases like this:

````
$msg = Session::pull($sessionName');
if (!empty($msg)) {
    return "<div class='alert alert-{$sessionName} alert-dismissable'>
            <button type='button' class='close' data-dismiss='alert' aria-hidden='true'>×</button>
            <h4><i class='fa fa-check'></i> success</h4>
          </div>";
}
````

will output:

````
<div class='alert alert-success alert-dismissable'>
    <button type='button' class='close' data-dismiss='alert' aria-hidden='true'>×</button>
    <h4><i class='fa fa-check'></i> success</h4>
</div>
````

daveismyname:
actually realised I was wrong!

PR #297, opened 2015-12-23
3.0 - Final Phase, stage continued

LuckyCyborg:
This pull-request introduce a demonstrative Module called **Demo**, which contains our Demo Controllers and complete the support for **Hooks Legacy/Compatibility Layer** in **\Nova\Events\Manager**, using, of course, the **Events API**. 

### Hooks Legacy/Compatibility Layer on Events Manager

To note that we do not recommend the use the **EventManger's HLCL**, instead we strongly suggest to port the existent code to the much more powerful native **Events API**, that **HLCL** existing only as quick&dirty solution to port the existing code to **Nova Framework**.

The usage of **HLCL** is simple. The Hooks should be registered using:

```php
EventManager::addHook("topnav", "\Modules\Broadband\Controllers\Broadband@nav");
```

And to run a Hook, we have:

```php
$args = '';

$topnav = EventManager::runHook("topnav", $args);
```

Also, there exists a Hook existing checking method, called **hasHook**, who, similar with **hasEvent**, return true when one or more Listeners are registered  to this Hook and can be used as:

```php
if(EventManager::hasHook("topnav")) {
    // do shiny things there about TOPNAV
}
```
To note that, in comparation with the original **Hooks** helper, being written on top of **Events API**, there are some subtle differences on **HLCL**, for example, there are no **Hook Domain**, literally a Hook's "where" now being  (part of) an **Event Name**. Then there will not be raised Exceptions when a new Hook "where" is registered. Also there is no need for a helper method **setHook($where)**.

Also, as difference to native Event deploying, the **HLCL** pass to the Listener Method, as argument, the  Result, in loop, similar with the original **Hooks Helper** style.

### Module demonstration : Demo

With this pull-request, our Demos are moved, like I said, in a Module called **Demo**, who have also the role to demonstrate **the Design of a Module** on the Framework, after we introduced our enhancements.


daveismyname:
Let me know when to merge.

LuckyCyborg:
@daveismyname OKay, Dave!

LuckyCyborg:
@daveismyname Please, merge this pull-request, being complete as objectives! :smile: 

I will open another devel pull-request in short time... :smirk:

PR #298, opened 2015-12-23
3.0 - Final Phase, stage continued #2

LuckyCyborg:
## WIP - DO NOT MERGE

### DIFF
--- a/app/Config/bootstrap.php
+++ b/app/Config/bootstrap.php
@@ -12,6 +12,7 @@ use Nova\Events\Manager as Events;
 use Nova\Net\Session;
 use Nova\Net\Router;
 use Nova\Config;
+use Nova\Logger;
 
 /**
  * Turn on output buffering.
@@ -25,6 +26,9 @@ require dirname(__FILE__).DS.'config.php';
 /**
  * Turn on custom error handling.
  */
+
+Logger::initialize();
+
 
 set_exception_handler('Nova\Logger::ExceptionHandler');
 set_error_handler('Nova\Logger::ErrorHandler');
--- a/app/Config/config.php
+++ b/app/Config/config.php
@@ -30,6 +30,13 @@ Config::set('routing', array(
 ));
 
 /**
+ * Profiler configuration
+ */
+Config::set('profiling', array(
+    'with_queries' => false
+));
+
+/**
  * All known Languages
  */
 Config::set('languages', array(
--- a/app/Config/config.php
+++ b/app/Config/config.php
@@ -16,6 +16,20 @@ use Nova\Events\Manager as Events;
 Config::set('timezone', 'Europe/Rome');
 
 /**
+ * Routing configuration
+ */
+Config::set('routing', array(
+    'patterns' => array(
+        //':hex'    => '[[:xdigit:]]+',
+        //':uuidV4' => '\w{8}-\w{4}-\w{4}-\w{4}-\w{12}'
+    ),
+    'auto_dispatch' => array(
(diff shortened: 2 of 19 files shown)


tomvlk:
@LuckyCyborg I think the catchall may be selective for get/post/put/delete etc. But the new default route is all.

LuckyCyborg:
@tomvlk The **Catch-All** principle is, well... to **Catch-All**, just like a **black-hole**. Any shiny Route shall be processed before or the the Request will end there.

I see only two uses for **Dave's Route**: in a **CMS** for routing dynamic slugged pages and as a "custom" Error404.

tomvlk:
@LuckyCyborg Some http methods may never return 404, or need to be handled different from others.

LuckyCyborg:
@tomvlk That's why I believe to have a utility the Dave's Route as custom Error404. I.e. in an **API**.

tomvlk:
@LuckyCyborg Would make it way better if you could just add a catchall for specific methods.

daveismyname:
You should be able to do that ie:

````
Router::get('(:all)' .....
````

LuckyCyborg:
@tomvlk Let's do not complicate the Routing without a visible utility...

Literally the Routing match a Route and its Method, so all the things are well defined. 

The Dave's Route make sense only as a black-hole where any un-matched Request arrive... :smirk:

tomvlk:
@daveismyname If that will be possible (or will stay possible) it's okay.

LuckyCyborg:
@daveismyname Yep, everyone is free and able to define Catch-All Routes per Method, as they wish...

I defined separately your special Route, to ensure that it is inserted last in the Routes list. In fact, if defined, is added to Routes list just before to launch the Dispatching.

This way, we can't "hide" the Routes defined by Modules, from Route order mistake.

LuckyCyborg:
@tomvlk It is just a special processing for a special Route. Nothing changed in rest.

tomvlk:
@LuckyCyborg Ok great!

LuckyCyborg:
@daveismyname BTW, Dave... 

For what is good to have patterns defined for **HEX**s an **UUID**s in Routing?

tomvlk:
@LuckyCyborg it's a good thing to have more patterns. But more generic. Such as double, float or something. Maybe even a-zA-Z too. 
But the uuid and hex are a bit specific. Maybe a good idea to make them dynamic in config?

LuckyCyborg:
@tomvlk Well, I already mussing to move the Routing defaults (aka **DEFAULT_CONTROLLER** and **DEFAULT_METHOD**) to **Config** options.

Maybe it is a better idea to leave by default only **':any', ':num', ':all'** paterns and move others in **Routing Config**uration options.

daveismyname:
it was something a PR added I don't use them myself and can't remember what their usage is.

LuckyCyborg:
@daveismyname So is safe to nuke them? They are really, but really specific, maybe somewhat useful in some API.

But always the programmer can **preg_match** the data in its landing method.

daveismyname:
Yeah go ahead.

LuckyCyborg:
@daveismyname I have to ask a favor to do for us. 

This night, after merging this pull-request, to create a **TAG**, something like **3.0-developer-preview** or something, with its shiny download files.

Tomorrow we have to present to our Manager a presentation/preview of current stage of **Nova Framework**.

I hope that we don't ask so much... :smile:

LuckyCyborg:
@daveismyname Okay, Dave! Please merge this pull-request, being complete as objectives!

PR #304, opened 2015-12-24
Adding add to Config class. To add values to a config value array.

tomvlk:
Just as the title say.
You can add values to an existing config value (array value).
This will allow modules to add a database link to the database config.

### DIFF
--- a/system/Config.php
+++ b/system/Config.php
@@ -10,18 +10,69 @@
 
 namespace Nova;
 
-
+/**
+ * Class Config, Will hold the config of the whole framework, including modules.
+ *
+ * @package Nova
+ */
 class Config {
-
+    /**
+     * @var array
+     */
     protected static $settings = array();
 
-
-    public static function get($key) {
+    /**
+     * Get value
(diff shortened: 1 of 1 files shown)


LuckyCyborg:
@tomvlk Goog Job, Tom! :+1: 

I susscribe to your propose! :smile: 

This time, you literally read my mind, this \Nova\Config::add was discussed at our Briefing as being necessary, i.e. for the Modules adding of Menu Items elements, or even in your scope, to extend the Database configuration.

Also, I see it being well guarded on errors. :+1:

tomvlk:
@LuckyCyborg Maybe I was there too :trollface: 
I missed it when I wanted to start on the demo stuff ;)

LuckyCyborg:
@tomvlk also, **Houston, we have a problem!**

You Services aren't **Modules aware**. We should talk about that problem and you should take action...

tomvlk:
@LuckyCyborg Look at the next pull request :trollface:

PR #308, opened 2015-12-24
PHPUnit Testing Base

tomvlk:
This will add the test base for letting PHPUnit run on every pull/commit.
Travis is fully configured, there is already 1 (messy) test case in place, it will test dispatching of the router.

But the focus will be more on testing db engines, because that's pretty important, and could be affected by several small changes.
And it's not only for testing the functionality right now, but more to maintain the quality of code that will not be touched in a while, but when it does, we are sure it still works (but not always 100% of coarse, you can never be...)..

Feel free to add comments, suggestions etc. 

@daveismyname Please add the project to https://travis-ci.org so it will start testing on push and pull requests. (By default you will get spam bombed if the 'build' (test) failes, you could turn this off I think).


tomvlk:
Forgot to say, When releasing you could completely ignore the tests folder, even delete it. 
Also the dependencies of the composer.json will be installed all, but when running **composer update --no-dev** or install/create project with the **--no-dev** flag on! 
It won't install all the PHPUnit dependencies.

tomvlk:
Btw, my travis results (from pull request source): https://travis-ci.org/tomvlk/php-framework/jobs/98723411

daveismyname:
Thank you!! I'm setting the account up now.

daveismyname:
I'll need you to take the lead on this, I'll be in America from 1st until the 18th with limited access to the internet.

daveismyname:
I've setup travis now. https://travis-ci.org/simple-mvc-framework/framework

tomvlk:
Great! I can see it working already! Lead on the testing stuff? That would be no problem but costs some time to make cases.

daveismyname:
Maybe it's not so important during the beta but should be checked once we feel everything is in place before releasing a RC

LuckyCyborg:
And who check, well.. the checks being correct? :smirk:

tomvlk:
Ye indeed. Anyway. I will work on the tests coming days! (And Christmas of course)

daveismyname:
Yes, important to spend some time with the family :)

LuckyCyborg:
@tomvlk And who check you, well... as writing correct tests?

LuckyCyborg:
@tomvlk And who establish the Rules of the Tests? You know, those "shall be..." :smirk:

nekomajin:
@tomvlk I'll support you as much as possible.

@LuckyCyborg We support each other. That's how it works in a community, right?

LuckyCyborg:
@tomvlk  and @nekomajin In the mean time, both look at this live test:

http://smvc3.giulianaeassociati.com/demos/database/service/basic/sqlite

You know what that test say? A saying of one my Team Leaders, from the Past:

If you want to learn to write Production Application, to remember that usually the Servers run Linux. Then, before to master PHP, is a must to master Linux, arriving to be fluent in Linux as using it as casual, for example,  watching your preferred movies and check the Mail.

Sadly, the server who host our live test website run, well.. as expected, on Linux.

tomvlk:
I'm almost sure it isn't me here. I'm Linux pro btw. Let me look at the blame view today. And we will see.

Issue #10, opened 2014-06-03
Propose update router like maclaw


andre2:
It is proposed to update the router as Maclaw to check route and halt when route exist.

Check out maclaw update: https://github.com/NoahBuscher/Macaw

PR #315, opened 2015-12-25
3.0 - Final Phase, stage continued #10

LuckyCyborg:
This pull-request improve the Modules awareness of **\Nova\Database\Manager** and complete the moving of Database Demos on the module Demo


tomvlk:
How about external tools, not in the \App and if they want to use the services, or declare services?

LuckyCyborg:
@tomvlk Well, I supposed that the Services should be in App or its Modules. Still, I do not see something special for tools who declare Services. Things should be usual. 

BTW, about what Tools you talk?

tomvlk:
@LuckyCyborg I ment more like if you have another module or something that extends Nova (different composer project) with another namespace start. But that would be very unusual so nvm.

LuckyCyborg:
@tomvlk Yeah, it is a valid question, yet somewhat out of Nova Framework's Context.

Still we have Applications with **Frontend/Backend** who use different namespaces:

```
\App\Backend\
\App\Frontend\
```

For resolving this situation, we use a constant defined on boot file (ie **/public/index.php**), called **APPNSPACE**, constant which is used then in Framework, i.e in Routing or Database Manager.

We do not proposed this for SMVC3/Nova for keeping the things simplified, but if Dave want it, there is no problem to make the changes required.

tomvlk:
@LuckyCyborg Btw, I found a bug. In the Core/Controller.php, line 85, you should create a new instance of \Exception. Could you fix that too here.

LuckyCyborg:
@tomvlk I seen it, thanks!

PR #317, opened 2015-12-26
3.0 - Final Phase, stage continued #11

LuckyCyborg:
Adding the method \Nova\Net\Router::matchedRoute and make the Callback optionally on default mode; useful for Testing


LuckyCyborg:
@daveismyname and @tomvlk I have a question for you:

**WHY** we need **LIMIT** on **UPDATE/DELETE** methods from Database Engine for MySQL?

To be honest, I never used LIMITs and I never seen them used, in this combination, in my 20 years length working with PHP. It is just another **multi-pass**?

To note that this thingie make us to copious duplicate the code in Database Engines.

PR #318, opened 2015-12-26
3.0 - Final Phase, stage continued #12

LuckyCyborg:
This pull request introduce the refactoring of Database Engines, to avoid the redundant code and to consolidate the structure.

In principle, the common used methods from **MySQL** and **SQLite** engines are moved in a base one, called **\Nova\Database\Engine\Base** and used as parent.

In other hand, some changes are introduced:

The **insertAll** method become **insertBatch** and now is totally independent from **insert** one, which lose its ability to manage the multi-inserts, to simplify the code and those multi-inserts being a MySQL thingy. Then, under **SQLite Engine**,  the method **insertBatch** will nuke the fridge (go Exception).

The methods **update** and **delete** lose the ability to establish **LIMIT**s as being a MySQL thingy and, considered by us, permitting developing bad habits on coding; as **UPDATE** and **DELETE** should be always atomic, with a very precise specified **WHERE**.

### PS. Our personal opinion about introducing Tests just for fun:

rudeboy

### DIFF
--- a/app/Config/config.php
+++ b/app/Config/config.php
@@ -97,7 +97,8 @@ Config::set('database', array(
             'username'      => 'root',
             'password'      => 'password',
             'fetch_method'  => \PDO::FETCH_OBJ, // Not required, default is OBJ.
-            'charset'       => 'utf8' // Not required, default and recommended is utf8.
+            'charset'       => 'utf8', // Not required, default and recommended is utf8.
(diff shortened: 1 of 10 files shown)


tomvlk:
I will update the tests after merge. Great work.

LuckyCyborg:
@tomvlk Thanks!

PR #320, opened 2015-12-27
3.0 - Final Phase, stage continued #14

LuckyCyborg:
Move to default fetching method PDO::FETCH_ASSOC as optimizing the database server bandwidth and introduce auto-configuration of fetching method in Services.

### DIFF
--- a/app/Config/config.php
+++ b/app/Config/config.php
@@ -96,7 +96,7 @@ Config::set('database', array(
             'database'      => 'dbname',
             'username'      => 'root',
             'password'      => 'password',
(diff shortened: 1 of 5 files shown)


daveismyname:
I prefer objects over arrays so when using this to use an object I would use FETCH_OBJ in my config class corrected? otherwise the default assoc would be used.

LuckyCyborg:
@daveismyname It is configured from Database configuration. 

You need only to put there **PDO::FETCH_OBJ**, instead of default **PDO::FETCH_ASSOC**

And the Services auto-switch the Engine in this mode.

daveismyname:
great.

Issue #15, opened 2014-06-19
Session pull function


avg206:
Hello, as I seen this item should be called pop function based on its functionality.

PR #321, opened 2015-12-27
3.0 - Final Phase, stage continued #15

LuckyCyborg:
Making the Database Services better citizens while auto-configuring the Engine's Fetch Method

### DIFF
--- a/system/Database/Service.php
+++ b/system/Database/Service.php
@@ -116,7 +116,7 @@ abstract class Service
      * @return false|Entity
      * @throws \Exception
      */
-    public function update($entity, $limit = 1)
+    public function update($entity)
     {
         $primaryValues = array();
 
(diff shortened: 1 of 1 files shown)


tomvlk:
Ehm. Services should always do a fetch class! With the entity as a class. This is incorrect

LuckyCyborg:
@tomvlk If I'm so wrong **why** this code pass your "tests" and the Database Demos works?

tomvlk:
Because the service tests are still not done in the 3.0-beta on here.

LuckyCyborg:
@tomvlk Sure, **BUT** why the Database Service Demos works?

tomvlk:
Because a class is also an object v.v. Demo doesn't uses entity itself, and is not checking it its instance of the entity. 
But when you are going to declare function in your entity and try to use it it doesn't work. 

Why not object? Entity can have functions like getName() that will concatenate first name with last name for example.

LuckyCyborg:
@tomvlk Okay, then also your Demos are incomplete. I understand. :smirk:

tomvlk:
Not directly incomplete. But there can be a toString for example that will provide a string notation for the car. 
But that's fully optional. Would be good in the demo I think.

LuckyCyborg:
@tomvlk I believe that would be better to have it in Demos.

tomvlk:
Ye. I can't do it right now. But will add it.

PR #326, opened 2015-12-28
3.0 - Final Phase, stage continued #19

LuckyCyborg:
Simplify the logic of Database Engines, permitting to configure the fetching return type with a single parameter instead of two.

The new parameter is **returnType**, with the following possible values:

**array** is equivalent with passing **PDO::FETCH_ARRAY**
**object** is equivalent with passing **PDO::FETCH_OBJ**
**className** is equivalent with passing **PDO::FETCH_CLASS** and to fetch a valid qualified a **Entity**, defined by a qualified Class Name (full namespace).

Also, in this pull-request is continued the work on **\App\Core\BaseModel** and new methods are implemented.

### DIFF
--- a/app/Config/config.php
+++ b/app/Config/config.php
@@ -91,22 +91,22 @@ Config::set('database', array(
     'default' => array(
         'engine' => 'mysql',
         'config' => array(
-            'host'          => 'localhost',
-            'port'          => 3306, // Not required, default is 3306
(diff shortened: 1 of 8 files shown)


LuckyCyborg:
@daveismyname Please merge this pull-request, as being completed as objectives!

Issue #25, opened 2014-08-01
[Feature]Auto Dispatch


volter9:
Why don't you add auto dispatch feature? That would be very handy.

Something like this:

```
public static function autoDispatch () {
    new \core\config;

    $uri = parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH);
    $uri = trim($uri, ' /');

    $parts = explode('/', $uri);

    $controller = $uri !== ''      && isset($parts[0])  ? $parts[0] : DEFAULT_CONTROLLER;
    $method     = $uri !== ''      && isset($parts[1])  ? $parts[1] : DEFAULT_METHOD;
    $args       = is_array($parts) && count($parts) > 2 ? array_slice($parts, 2) : array();

    // Check for file
    if (!file_exists('app/controllers/' . $controller . '.php')) {
        // Process 404 here 
        self::processRoute(self::$error_callback, 'No routes found.');

        return;
    }

    $controller = '\controllers\\' . $controller;
    $c = new $controller;

    if (method_exists($c, $method)) {
        $c->$method($args);
    }
}
```

This method would find controller, method and arguments for controller.
Similar to CI's routing: `/controller/method/arg1/arg2/and/so/on`

Where DEFAULT_CONTROLLER and DEFAULT_METHOD are the constants that do what they sounds. self::processRoute is static method that I created, because you guys too much of repetition in dispatch method (same lines three times!).

What do you think?

P.S.: couldn't find how to add a question label :confused:

PR #335, opened 2015-12-29
3.0 - Final Phase, stage continued #25

LuckyCyborg:
This pull-request contains the required support, implemented into Database Engines and BaseModel, to create a checking of existent Columns into Database Table and to compare with passed Data, then to smart use only the fields who have an equivalent Column into current Table. What is good for?

First of all, a more relaxed way to pass the Data to **BaseModel**, without specially selecting the required fields and unsetting the rest of them. Also, avoid the trying to write to Engines using unavailable fields of information, usually with disastrous results.  

Combined with BaseModel's ability to pre-process the Data, using one of its own Events and its own Event Management, can permit a useful simplification on writing strong but simple code.

Practically, a typically flow will be: the POST Data arrive to Validation, who check the values, using its presets, eventually filtering the Data content, then the resulted information array can be passed directly to the Model (fi being a BaseModel child), which will select only the useful parts of information, process them and finally passing to current Engine.

Alternative, we can configure into Model an array called **$fields**, who contains the known Table Column name, case when BaseModel will not try to auto-detect the values using a supplementary Query, event it is about one for instance, first when a method who need to map the information to the known columns is called.

Sure, one can argue that will be better to integrate the Validation directly into (Base)Model, to ensure that everything is well filtered and validated, speculating the existence of different methods for INSERT and UPDATE, the typical cases where the Validation is need.

While possible this combination, and even used by some Big Houses, i.e. CakePHP, we believe that is better to keep separated those two parts, Validation and Model, for a better control of the flow.

Finally, for those who wonder about what is our **BaseModel**, let's say that is somewhat similar as abilities with those Base Models used in the **CodeIgniter** World.

Or, in other words, the new **Nova Framework**, with its Routing and this BaseModel, combined with the Controller/View duet, is something like CodeIgniter as Models and like CakePHP as Controllers/Views style, with a bit of Laravel, but much more simple as structure and learning path.

We hope that our Design will be liked by everyone. :smirk:


LuckyCyborg:
@daveismyname Please merge this pull-request, as being complete as objectives!

PR #355, opened 2016-01-02
3.0 - Doctrine DBAL based support #1

LuckyCyborg:
Initial implementation.

### DIFF
--- a/app/Config/config.example.php
+++ b/app/Config/config.example.php
@@ -96,8 +96,8 @@ Config::set('database', array(
             'database'    => 'dbname',
             'user'        => 'root',
             'password'    => 'password',
-            'return_type' => 'array',     // Not required, default is 'array'.
             'charset'     => 'utf8',      // Not required, default and recommended is utf8.
+            'return_type' => 'array',     // Not required, default is 'array'.
             'compress'    => false        // Changing to true will hugely improve the persormance on remote servers.
         )
     ),
--- a/app/Config/config.example.php
+++ b/app/Config/config.example.php
@@ -89,6 +89,7 @@ Config::set('emailer', array(
 Config::set('database', array(
     'default' => array(
         'engine' => 'mysql',
+        'driver'  => 'pdo_mysql',
         'config' => array(
             'host'        => 'localhost',
             'port'        => 3306,        // Not required, default is 3306
@@ -103,6 +104,7 @@ Config::set('database', array(
     /** Extra connections can be added here, some examples: */
     'sqlite' => array(
         'engine' => 'sqlite',
+        'driver'  => 'pdo_sqlite',
         'config' => array(
             'file'         => 'database.sqlite',
             'return_type'  => 'object' // Not required, default is 'array'.
--- a/app/Config/config.example.php
+++ b/app/Config/config.example.php
@@ -94,7 +94,7 @@ Config::set('database', array(
             'host'        => 'localhost',
             'port'        => 3306,        // Not required, default is 3306
             'database'    => 'dbname',
-            'username'    => 'root',
+            'user'        => 'root',
             'password'    => 'password',
             'return_type' => 'array',     // Not required, default is 'array'.
             'charset'     => 'utf8',      // Not required, default and recommended is utf8.
@@ -106,7 +106,7 @@ Config::set('database', array(
         'engine' => 'sqlite',
         'driver'  => 'pdo_sqlite',
         'config' => array(
-            'file'         => 'database.sqlite',
+            'path'         => BASEPATH .'storage/persistent/database.sqlite',
             'return_type'  => 'object' // Not required, default is 'array'.
         )
     )
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -43,7 +43,7 @@ Config::set('database', array(
             'host'        => 'localhost',
             'port'        => 3306,     // Not required, default is 3306
             'database'    => 'testdb1',
-            'username'    => 'root',
+            'user'        => 'root',
             'password'    => '',
             'return_type' => 'object', // Not required, default is 'array'.
             'charset'     => 'utf8'    // Not required, default and recommended is utf8.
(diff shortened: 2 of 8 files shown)


tomvlk:
@LuckyCyborg Please leave the current engines as it is until the tests have been replaced and refactored for the new DBAL and services.

LuckyCyborg:
@tomvlk Worry not! I will implement our little **ORM** in, well... **\Nova\ORM** :smirk:

tomvlk:
Weowowwowowo take it easy :laughing: keep some work for me please.

LuckyCyborg:
@tomvlk 

> Weowowwowowo take it easy :laughing: keep some work for me please. 

Correction: **We** will implement our little **ORM** in, well.. **\Nova\ORM**. That's better? :smile:

tomvlk:
Good enough haha

LuckyCyborg:
@tomvlk Let's say that on the (mini-)**ORM** part we will make the skeleton and you will polish it and complete. You are in? :smirk:

tomvlk:
@LuckyCyborg Great!

LuckyCyborg:
@tomvlk Okay, Tom!

But, I expect you comments and your hands into **DBAL** thing, also... :smirk:

tomvlk:
@LuckyCyborg Of course, Non ti preoccupare

tomvlk:
@LuckyCyborg Btw, what about to make this:
**\Nova\Database\Entity** for using in generic entities.
**\Nova\ORM\Entity** extends **\Nova\Database\Entity** for use with ORM methods, like save, update, find, etc.

LuckyCyborg:
@tomvlk Good one! Nothing against.

But, lets arrive to ORM... :smirk: 

You, meantime, do a eagle-flight over the simple ORM implementations. I wait to give me some opinions and suggestion on that theme... :smile:

LuckyCyborg:
@tomvlk And there we go!

Right now, in theory we are ready to return a **\Nova\DBAL\Connection** from DBAL Manager...

tomvlk:
## ORM Idea
#### Entity
The entities need to extend **\Nova\ORM\Entity** and use Annotations (customized annotations we create in our ORM\Annotations\ namespace) for table name, column names and database link name.

Sample entity:
```php
/**
 * Car Entity
 * @Table(name="car")
 **/
class Car extends Entity 
{
  /**
   * @var string
   * @Primary
   * @Column(type="int")
   **/
  public $carid;

  /**
   * @var string
   * @Column(type="string")
   **/
  public $make;

  /**
   * @var string
   * @Column(type="string")
   **/
  public $model;

  /**
   * @var double
   * @Column(type="double")
   **/
  public $costs;

  /**
   * @var string
   * @Column(type="string")
   * @Null
   **/
  public $type;
}
```
Or we could stick with the protected Entity variables for determinating table name, primary key etc.


#### ORM Methods
I was thinking of the following methods, remember, $car is an car already fetched from the database:
```php
$car->save(); // Save the entity, or update when already exists.
$car->update(); // Only update, will not create.
$car->create(); // Only insert, no update!
$car->delete(); // Will delete it from the db

$select = Car::find();
$car = $select->one()->where(array('carid' => 1))->fetch();
// $car wil be a single Car instance.

$allCars = $select->all()->fetch();
// $allCars will be an array with Car instances
```

LuckyCyborg:
@tomvlk You got right the **Entity** structure, with its **Annotations**. See that you figured immediately how to use the shiny things coming with **Doctrine DBAL** ? ;smirk:

About the composite select methods, lets see what the **QueryBuilder** permit...

tomvlk:
@LuckyCyborg Ye, the select isn't really been done. But that is dependent on the QueryBuilder indeed.
I will start on the **Entity** in the upcoming few days. But for sure around Monday late and Tuesday.

LuckyCyborg:
@tomvlk The **select** method is there. Look on #356

tomvlk:
@LuckyCyborg Just saw it! I think we still need to test our own **Connection** we extend from Doctrine, as we change it a bit and will add some helpfull methods and tools into it. Like the counter and select.

LuckyCyborg:
@tomvlk Of course we need to test it. :smirk: 

I will implement late today an example/demo/test for this **DBAL** and you are free to make your Tests, too... :smile:

tomvlk:
@LuckyCyborg Keep it easy, don't over rush it. We don't even know what Dave thinks about this, But I'm sure he gonna love the Doctrine QueryBuilder for sure!

tomvlk:
@LuckyCyborg Maybe an idea to use our manager (dbal manager) to get a query builder directly, without having to go to the getConnection and then use the getQueryBuilder. Just for easy.

Btw: See manager, line nr 19, typo!

LuckyCyborg:
@tomvlk 

> Btw: See manager, line nr 19, typo!

Enlighten me, what typo?

tomvlk:
https://github.com/simple-mvc-framework/framework/blob/3.0-beta/system/DBAL/Manager.php#L19
nstances  -> instances

LuckyCyborg:
@tomvlk Thanks! I seen it!

LuckyCyborg:
@tomvlk 

> Maybe an idea to use our manager (dbal manager) to get a query builder directly, without having to go to the getConnection and then use the getQueryBuilder. Just for easy.

Hmmm... But, practically the things are transparently and are supposed to be managed in **Model**.

Remember? It is supposed that when you use a **Model** (or **Entity** ?) you should already have a well configured and initiated **Database Connection**.

tomvlk:
Yep, But in case anyone doesn't want to use the Model/Entity (ORM). And still want the QueryBuilder.

LuckyCyborg:
@tomvlk 

> We don't even know what Dave thinks about this, But I'm sure he gonna love the Doctrine QueryBuilder for sure!

Considering that he loved the **BaseModel**'s mini-QueryBuilder methods, I bet that he will be very satisfied to have one of most powerful QueryBuilder existent.

In other hand, if someone want to use the QueryBuilder part without Models/Entities, have just to get the **Connection** instance.

BUT, remember, Nova Framework is a **MVC** one, then is supposed to use Models there.

tomvlk:
@LuckyCyborg Ye you are right about the models. But I mean why not create it for usage for in events for example, just small things.

Entity is normally supposed to be an table that will be used, the ORM.

LuckyCyborg:
@tomvlk The Events aren't supposed to interact directly with **DBAL**. 

It is supposed those Events to invoke (live) Controllers or even bare Classes methods, who will call Model to do their jobs... :smirk:

PR #377, opened 2016-01-04
Ensure that we set the State in a valid Entity retrieved

LuckyCyborg:
Like the title say.


tomvlk:
Thanks for the hot patch.

LuckyCyborg:
@tomvlk No problem! :smirk:

PR #398, opened 2016-01-06
3.0 a bunch of fixes

nekomajin:
- correction of {min} and {max} values
- fix typo: Tshould -> should
- fix definition: images we measure in pixels, files in bytes
- corrected sentence structure: "The image does must have a ratio..." -> "does must have" make sense?

@LuckyCyborg Please correct me if I'm wrong. If everything is right I will send the german translation.


LuckyCyborg:
@nekomajin Thanks for your efforts!

And Yes, looks OKay your strings. To note that what I made was to copy&paste the messages from vendor;s Validator Rules, needing to integrate the Validation into our **INTL** Envinroment.

Specially I checked if changing the label to {max} is supposed to work in the corespondent **Validation Rule**, and everything looks OKay.

PR #404, opened 2016-01-06
Changed valid domain name error

Kingorgg:
Same as title.

### DIFF
--- a/system/Language/en/messages.php
+++ b/system/Language/en/messages.php
@@ -39,7 +39,7 @@ return array (
   'This input should have at least {min} characters' => '',
   'This input should have less than {max} characters' => '',
   'This input should not match the regular expression {pattern}' => '',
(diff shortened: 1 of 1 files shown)


Kingorgg:
Sorry, Please Ignore, I'll Fork and make a proper commit, instead of using the browser.

LuckyCyborg:
@Kingorgg Looks like you try to modify the Translation Strings.

That's a thing that we should never do! 

First of all, they are just reference strings, who live somewhere, in reality. Modifying that will make unavailable. 

In other hand, there is an Language Strings Updater, which will friendly nuke the non-existent strings, then your work will gone on first update.

What you have to do? Please find the suspect string into Nova's structure, not in translation file, and change it.

Kingorgg:
@LuckyCyborg: I see, Thanks for letting me know, I won't touch it. Sorry!

PR #406, opened 2016-01-06
3.0 small changes

nekomajin:
- Point added at the end of a sentence, to unify.
- fix typo: "IDs should be an array if primary key values!" => if -> of


LuckyCyborg:
Thanks!

PR #411, opened 2016-01-07
3.0 update german language file

nekomajin:


LuckyCyborg:
@nekomajin Thanks! 

Keeping this pace, the Nova Framework will speak German like it was born in Frankfurt... :smirk:

PR #443, opened 2016-01-09
Revert "3.0 - Several fixes in Nova DBAL"

LuckyCyborg:
Reverts simple-mvc-framework/framework#442

### DIFF
--- a/system/Database/Connection.php
+++ b/system/Database/Connection.php
@@ -177,29 +177,24 @@ abstract class Connection extends PDO
 
         // Bind the key and values (only if given).
         foreach ($params as $key => $value) {
-            $bindKey = $prefix .$key;
-            if (is_int($key)) {
-                $bindKey = $key+1;
-            }
-
             if(isset($paramTypes[$key])) {
-                $statement->bindValue($bindKey, $value, $paramTypes[$key]);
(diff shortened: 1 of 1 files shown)


LuckyCyborg:
What you do not understand that you cannot rely into this numeric keys?

Look there:

```php
        // Bind fields
        $this->bindParams($stmt, $data, $paramTypes, ':field_');

        // Bind values
        $this->bindParams($stmt, $bindParams, $paramTypes);
```

tomvlk:
@LuckyCyborg Please don't just force it! You should really consider talking a bit more and maybe understand it.

tomvlk:
@LuckyCyborg Query builder can! Because you prepare it and it will be numeric all the time!

LuckyCyborg:
@tomvlk Then use directly the QueryBuilder and do not touch the **Connection** introducing strange bugs!

tomvlk:
@LuckyCyborg Please stop the childish Adrian here.

LuckyCyborg:
@tomvlk I'm not childish. Look into code and you will see that the **bindParam** should be index safe to use numeric bindings.

Because it can be called several times in a single method.

tomvlk:
@LuckyCyborg Ok, please make it so you can use question mark bindings!

tomvlk:
@LuckyCyborg But not revert right now.

PR #457, opened 2016-01-10
Fix the Tests

LuckyCyborg:
And there we go! 

It's about *my favorite* thing: **Fix the Tests**.

@daveismyname BTW, Dave. You really think that you need those Tests who manage to double the work of your programmers, just for fun?

### DIFF
--- a/scripts/smvc_members.sql
+++ b/scripts/smvc_members.sql
@@ -23,20 +23,20 @@ SET time_zone = "+00:00";
 -- --------------------------------------------------------
 
 --
--- Table structure for table `smvc_members`
+-- Table structure for table `nova_members`
 --
 
-CREATE TABLE IF NOT EXISTS `smvc_members` (
+CREATE TABLE IF NOT EXISTS `nova_members` (
   `id` int(11) unsigned NOT NULL,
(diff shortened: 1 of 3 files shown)


LuckyCyborg:
@daveismyname I will continue tomorrow. It is 3AM there. 

Meantime the Framework's build fail because you changed something in a example...

That's the Tests fun! :smirk:

daveismyname:
on the other hand it alerts us to silly mistakes! I will correct.

daveismyname:
I obviously don't get what's going on, I chanced smvc to nova in constants then you changed the testing_mysql.sql to use nova instead, the script drops the table if exists then creates the table.

Yes in the tests is failing as the table does not exist. What am I missing?

tomvlk:
I will look into it today. 
And I'm pretty sure tests help. I already saw lucky correct himself because a test failed and he saw that he made a mistake :trollface:

tomvlk:
It was the SQLite database and there was a bug in one of the demo files that caused the tests to fail, so the tests were right, found a bug, and the table in the SQLite db wasn't changed.
I just merged a pull that fixes this.

PR #458, opened 2016-01-10
3.0 - Fix in demo route + Renamed table prefix for tests

tomvlk:


LuckyCyborg:
@tomvlk You known what is amazing?

I'm look into **app/Modules/Demo/Config/routes.php** and right as I remembered, my current tree have that **use**, as supposed to be.

Also look there:

https://github.com/LuckyCyborg/simple-mvc-framework/blob/3.0-beta/app/Modules/Demo/Config/routes.php

Both have the right use, as:

```php
use Nova\Net\Router;
```

Then, amaze me! How you arrived to that "bug" and patch?

tomvlk:
@LuckyCyborg Please look better. We can blame @daveismyname with his commit: https://github.com/simple-mvc-framework/framework/commit/313b97a5632b27e48cdc21314709c5c2071831dc

LuckyCyborg:
@tomvlk Ha!? Thanks! :+1: 

Looks like that happened last night, after i quit. :smirk: 

BTW, you ORM is kicking and well, after I written for it a special **Nova\Database\Connection** extension/wrapper who try the best to use unnamed parameters binding, instead of named ones, for some essential methods.

If you need more, just tell, I will improve further that specialization for Connection. :smirk:

daveismyname:
Yes my fault I must have removed the line instead of removing the extra blank line!

Issue #478, opened 2016-01-11
Native database


jimgwhit:
@daveismyname and @LuckyCyborg have you seen the pull request on the fluentpdo github site?  
If we are not using doctrine dbal, could we not use these packages either?  If everything is written by dave or @LuckyCyborg or @tomvlk then updating the framework will be easier.  As of now I have to look for updates in nova and fluentpdo to stay up to date.  I know you want to move forward and finalize, but give some thought to it please.

LuckyCyborg:
@jimgwhit about what pull-request you talk, Jim?

jimgwhit:
The ones on their site.

jimgwhit:
@LuckyCyborg ok I am now trying nova again without doctrine dbal, but using the native db.  I am getting a table showing, but no data, here is model, how to change to work?
```
<?php

namespace App\Models;

use Nova\Net\Session;

class PetModel extends \App\Core\BaseModel
{

    public function __construct()
    {
        parent::__construct();
    }
    
    public function petCount($petsearch = "")
    {
        $petsearch = $petsearch . "%";
        //echo "you are here".Session::get('owner');
        //$ownerid = '1';
        //echo $this->_db->kselect('SELECT count(petid) as total FROM ' . PREFIX . 'pets WHERE petname LIKE :search',
        //array('search' => $petsearch));
        if (Session::get('owner') == '1') {
            echo "you are here" . Session::get('owner');
            $result = $this->db->select('SELECT COUNT(petid) as total FROM ' . DB_PREFIX . 'pets WHERE petname LIKE :search', array('search' => $petsearch));
            return $result[0]->total;

            //return $this->_db->kselect('SELECT count(petid) as total FROM ' . PREFIX . 'pets WHERE petname LIKE :search', 
            //array('search' => $petsearch));	
        } else {
            $result = $this->db->select('SELECT count(petid) as total FROM ' . DB_PREFIX . 'pets WHERE petname LIKE :search AND ownerid =
		:ownerid', array('search' => $petsearch, 'ownerid' => Session::get('owner')));
        }
        return $result[0]->total;

        //good return $this->_db->kselect('SELECT count(petid) as total FROM ' . PREFIX . 'pets WHERE petname LIKE :search',
        // array('search' => $petsearch));	
    }

    // good public function get_pets($limit="", $petsearch = ""){
    public function getPets($offset = "", $rowsperpage = "", $petsearch = "")
    {
[...]

PR #606, opened 2016-02-10
Update Statement.php

pandory-network:


tomvlk:
Nice bugfix! Thanks for providing this fix, this is most certainly fixing the issue in #478.

PR #607, opened 2016-02-15
Update Form.php

jcroucher:
Did these changes for my own project. Thought they may be of use for others.

- Changed the select place holder from "Select" to an optional parameter
- Added a new type that allows for automatically generating numbered drop down lists
- Added functionality to automatically check the correct radio button based on the value
- Added function to generate label wrapper for fields


daveismyname:
Thanks!

PR #609, opened 2016-02-19
return current month' number of days if month value is lower than 1 or bigger than 12

avenirer:


avenirer:
May I know the reason why not accepting this pull request? Thank you.

daveismyname:
I merged your more recent PR or have I got this wrong?

avenirer:
actually there are two PR's on same file... with different modifications.

avenirer:
Sorry about that, but I didn't want to do two different changes in the same pull request. I guessed I messed that up. Next time will be more careful.

daveismyname:
actually I should have merged the first one first, doing it back to front has not caused conflicts, I'll correct that later or if you have time can you update the PR.

PR #610, opened 2016-02-19
No need to check year in there...

avenirer:
Why would you need to check the year in there?


daveismyname:
Thanks your right $year is not needed.

PR #616, opened 2016-02-27
Revert "Applied fixes from StyleCI"

daveismyname:
Reverts simple-mvc-framework/framework#614

### DIFF
--- a/app/Controllers/Welcome.php
+++ b/app/Controllers/Welcome.php
@@ -1,25 +1,26 @@
 <?php
 /**
- * Welcome controller.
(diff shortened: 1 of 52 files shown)


LuckyCyborg:
I wonder why you play with this tools.

You believe that your target audience, who demonstrated that can barely understand how it works and analyze that ol'good Database Helper from 2.2, will understand the noble art of those thingies?

daveismyname:
the users are at various levels, the idea was if anything was not PSR compliant it would pick it up clearly needs more care.

PR #617, opened 2016-03-04
Update Gump.php

woodsy1:
Added in alpha space method.


woodsy1:
simple fix for alpha space in gump helper

PR #622, opened 2016-03-10
Automatically tableize the Template URLs, to properly be counterpart …

LuckyCyborg:
…to the new Routing's File Serving

The basic idea is that a Template path like:

```
nova\app\Templates\NiceTheme/
```

Will become:

```
/templates/nice_theme/
```

The actual Routing known how to interpret this URI path and to look in the right path, transforming that path back into CamelCase.


daveismyname:
nice!

PR #632, opened 2016-03-14
Fix the Routing & Co.

LuckyCyborg:
#### Rationale

There is no such thing like **lang** session right now, then the **Language** helper fail to load the proper file.

So, that patch will fix the problem described into my latest post.

All the best!


daveismyname:
thank you.

PR #639, opened 2016-03-17
allow loading and retrieval of multiple language terms

avenirer:
This will allow loading more than one language. Also it allows the use of a different language for same term when retrieving it.

### DIFF
--- a/app/Core/Language.php
+++ b/app/Core/Language.php
@@ -36,7 +36,7 @@ class Language
         /* check if is readable */
         if (is_readable($file)) {
             /* require file */
-            $this->array = include $file;
+            $this->array[$code] = include $file;
         } else {
             /* display error */
             echo Error::display("Could not load language file '$code/$name.php'");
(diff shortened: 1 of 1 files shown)


LuckyCyborg:
I would like to raise your attention to another branch existing there:

https://github.com/simple-mvc-framework/framework/tree/3.0-nova

That branch is very different from what you see into the main branches, including a strong Language management, a small DBAL, a Laravel-esque ORM and even a CodeIgniter-ish BaseModel. 

Considering your experience with CodeIgniter, will be nice to take a look, for example, on App\Core\BaseModel.

Sure, feel free to make any suggestions for improvements, considering that that branch wants to be a small framework (then more complex) than the current one.

avenirer:
Thank you, @LuckyCyborg , but I still don't understand... what is Nova? Is it what will be called simplemvcframework v3? Or is it going to be a "more complex" version of a main simplemvcframework v3? Also, is there a release date established for that?

LuckyCyborg:
@avenirer 

> what is Nova?

**Nova Framework** is the new name/branding of **Simple MVC Framework**, supposed to follow with the **v3**.

To note that this branch (3.0-nova) doesn't represent the future SMVC, which you will find into branch **3.0**.

Historically, the 3.0-nova was the previous **3.0-beta** branch, and supposed to be the next SMVC version (aka Nova). In the same time, there was the start of my collaboration with SMVC team, being instructed to improve the SMVC, until is able to drive a complex e-learning portal, for its use on some of our projects (yeah, as in a job task).

Well, while us we need a small and fast framework, yet powerful enough to do complex tasks, then I written what you see, looks like the final result was far away from the target audience of SMVC, as in very simple MVC (for beginners?), even we used to strive to offer back compatibility layers, resulting in the end, with the demise of the current code and the restarting of 3.0 development.

Actually, the future v3 (Nova?), is what you see into branch **3.0**, practically, from a MVC structure POV, being the version **2.2**, but driven by the Routing from 3.0-nova and reorganized to permit the use of public webroot.

In the mean time, we continue our development into branch 3.0-nova, after all, that thing being written for (our) specific needs, but you can consider this branch as an independent framework and/or a personal pet project, which happen to have common elements with SMVC.

Still, there is no release date for this branch, just it will be developed as is.

So, if you like to play with a small and fast framework, there it is.

PR #661, opened 2016-03-22
add server status code for redirect

avenirer:
maybe someone needs it...


jimgwhit:
@avenirer could you explain what that status code is for and how one would use it please.

LuckyCyborg:
@jimgwhit If you don't know about that thing, it is safely to ignore it.

Let's say it is for those who want to fine tuning their redirections.

PR #701, opened 2016-04-05
Improve the URI detection and fix #700

### LINKED ISSUE #700: Problem with routes
Hello,

I try NOVA, and have problem routes when use GET.

When try buid a blog with Screen cast tutorial, all is ok but when use Helper Paginator, i can´t get it work.
I have controller Blog with 2 functions index and post, in index function i use Helper paginator and paginator apear correctly but when try go to another page i receive this error : Missing argument 1 for App\Controllers\Blog::post() in C:\xampp\htdocs\www\NOVA_FRAMEWORK\app\Controllers\Blog.php on line 51.

Line 51 is function post and seems router try to load function post and not function index for next posts pages.

In attachment test files.
[NOVA_FRAMEWORK.zip](https://github.com/nova-framework/framework/files/203516/NOVA_FRAMEWORK.zip)

LuckyCyborg:
This PR improve the URI detection into **Url::detectUri()** and fix the Issue #700

### DIFF
--- a/system/Helpers/Url.php
+++ b/system/Helpers/Url.php
@@ -53,11 +53,11 @@ class Url
             $requestUri = substr($requestUri, strlen($pathName));
         }
 
-        if (($requestUri == '/') || empty($requestUri)) {
+        $uri = parse_url($requestUri, PHP_URL_PATH);
+
+        if (($uri == '/') || empty($uri)) {
             return '/';
         }
-
-        $uri = parse_url($requestUri, PHP_URL_PATH);
 
         return str_replace(array('//', '../'), '/', ltrim($uri, '/'));
     }
--- a/system/Helpers/Url.php
+++ b/system/Helpers/Url.php
@@ -53,13 +53,14 @@ class Url
             $requestUri = substr($requestUri, strlen($pathName));
         }
 
-        $uri = parse_url($requestUri, PHP_URL_PATH);
+        $uri = parse_url(ltrim($requestUri, '/'), PHP_URL_PATH);
 
-        if (($uri == '/') || empty($uri)) {
-            return '/';
+        if (! empty($uri)) {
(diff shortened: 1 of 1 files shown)


jimgwhit:
@daveismyname and @LuckyCyborg in my install I am not having any issues with pagination, I have this
```
$data['pageLinks'] = $pages->pageLinks('?', '&psch=' . $dogsearch . '&aval=' . $aval);
```
So I have addional arguments being passed with GET.
Are you sure the OP in issue #700  was passing args to the paginator correctly.

LuckyCyborg:
@jimgwhit Thanks for your concerns but, yes, I can confirm that there is a Routing issue, in the particular case of empty URI with query parameters, and it can be reproduced easily under bare installation, for example the following URL will generate an 404, as in attached screenshot:

```
http://www.novaframwork.dev/?p=2
```

While the use of Pagination under Homepage triggered it, the issue appear because the **Url::detectUri()** return an empty string aka **""**, in this particular case, instead of the supposed **"/"** and the workaround is rather simple, as in moving up the URI cleanup, before evaluating the URI emptiness.  

snapshot4

jimgwhit:
@LuckyCyborg was it working in my installation because I was passing more parameters in the query string? I had page and two others but the two others were empty.

LuckyCyborg:
@jimgwhit The issue is triggered when you have an empty URI (for example, the URL of homepage) and you have query parameters. In other cases, when you have an URI defined (e.g. "/dogs/show?p=2"), everything works OK, like you supposed.

Like I said, it is about a niche case, that's why was more difficult to be seen and catch this bug.

In other hand, I'm glad that I hunted it down. Call me perfectionist, if you want.

jimgwhit:
@LuckyCyborg thanks for your explanation.

daveismyname:
Thanks! apologies for the delay having a very busy week.

PR #722, opened 2016-04-12
Overall fixes and logic improvements

LuckyCyborg:
This PR introduce overall fixes and optimizations, also try to improve the logic.

Notable, parts of the logic are moved to bootstrap stage to ensure the initialization steps, a small but very useful collection of site wide functions is introduced, and not as last, **Core\Alias** is renamed as **Aliases** and its logic is adjusted to be into Laravel style.

To be honest, the **Core\Aliases** is the first candidate to have a Configuration support, using a Config class like one present into 3.0-nova, as to permit the simple modifications of the class aliases...

But I leave to your decision if we introduce a dynamic configuration support; after all, not everything can be configured with some defines...

And a last question: we really need to enclose the configuration from app/Config.php in a class? Everything can work similar (while loaded by bootstrap), if there is just a bunch of defines and the error handling setup.


daveismyname:
With that functions file I'm thinking of moving the methods within system/Helpers/Data.php to the functions file, so data can be inspected/changed easily such as doing a quick pr:

````
pr($data);
````

to have a fast way to have:

````
echo "<pre>";
print_r($data);
echo "</pre>";
````

**Comment by LuckyCyborg** [comment]
@daveismyname Good point!

In fact, we can wrap the Helpers\Data methods into (some) functions...

Which candidates are, in your opinion?

**Comment by daveismyname** [comment]
I was thinking have the functions in the functions file instead of having the data helper at all.

**Comment by LuckyCyborg** [comment]
@daveismyname  Okay then.

And what you think about transforming the **app/Config.php** in a, well... configuration file and introducing a small dynamic configuration class, like:

```php
class Config
{
    /**
     * @var array
     */
    protected static $settings = array();

    /**
     * Get value
     * @param string $key
     * @return mixed|null
     */
    public static function get($key)
    {
        return isset(self::$settings[$key]) ? self::$settings[$key] : null;
    }

    /**
     * Set value
     * @param string $key
     * @param mixed $value
     */
    public static function set($key, $value)
    {
        self::$settings[$key] = $value;
    }
}
```

For example, with that Config we can move the Aliases configuration to (non-class) app/Config.php

daveismyname:
what would that look like?

LuckyCyborg:
```php
/**
 * Configuration constants and options.
 */

/**
 * Define the complete site URL.
 */
define('SITEURL', 'http://www.novaframework.dev/');

/**
 * Define relative base path.
 */
define('DIR', '/');

/**
 * Set the Application Router.
 */
// Default Routing
define('APPROUTER', '\Core\Router');
// Classic Routing
// define('APPROUTER', '\Core\ClassicRouter');

/**
 * Set default controller and method for legacy calls.
 */
define('DEFAULT_CONTROLLER', 'Welcome');
define('DEFAULT_METHOD', 'index');

/**
 * Set the default template.
 */
define('TEMPLATE', 'Default');

/**
 * Set a default language.
 */
define('LANGUAGE_CODE', 'En');

//database details ONLY NEEDED IF USING A DATABASE

/**
 * Database engine default is mysql.
 */
define('DB_TYPE', 'mysql');

/**
 * Database host default is localhost.
 */
define('DB_HOST', 'localhost');

/**
 * Database name.
 */
define('DB_NAME', 'dbname');

/**
 * Database username.
 */
define('DB_USER', 'root');

/**
 * Database password.
 */
define('DB_PASS', 'password');

/**
 * PREFER to be used in database calls default is smvc_
 */
define('PREFIX', 'nova_');

/**
 * Set prefix for sessions.
 */
define('SESSION_PREFIX', 'nova_');

/**
 * Optional create a constant for the name of the site.
 */
define('SITETITLE', 'Nova V3.0');

/**
 * Optionall set a site email address.
 */
// define('SITEEMAIL', 'email@domain.com');

/**
 * Turn on custom error handling.
 */
define('APPLOGGER', 'Core\Logger');

/**
 * Set timezone.
 */
define('DEFAULT_TIMEZONE', 'Europe/Rome');

Config::set('class_aliases', array(
            'Errors'        => '\Core\Error',
            'Language'      => '\Core\Language',
            'Mail'          => '\Helpers\PhpMailer\Mail',
            'Assets'        => '\Helpers\Assets',
            'Arr'           => '\Helpers\Arr',
            'Cookie'        => '\Helpers\Cookie',
            'Csrf'          => '\Helpers\Csrf',
            'Data'          => '\Helpers\Data',
            'Date'          => '\Helpers\Date',
            'Document'      => '\Helpers\Document',
            'Form'          => '\Helpers\Form',
            'Ftp'           => '\Helpers\Ftp',
            'GeoCode'       => '\Helpers\GeoCode',
            'Hooks'         => '\Helpers\Hooks',
            'Inflector'     => '\Helpers\Inflector',
            'Number'        => '\Helpers\Number',
            'Paginator'     => '\Helpers\Paginator',
            'Password'      => '\Helpers\Password',
            'RainCaptcha'   => '\Helpers\RainCaptcha',
            'Request'       => '\Helpers\Request',
            'ReservedWords' => '\Helpers\ReservedWords',
            'Response'      => '\Helpers\Response',
            'Session'       => '\Helpers\Session',
            'SimpleCurl'    => '\Helpers\SimpleCurl',
            'TableBuilder'  => '\Helpers\TableBuilder',
            'Tags'          => '\Helpers\Tags',
            'Url'           => '\Helpers\Url'
));

daveismyname:
OK the only part that I jumps out at me is:

````
define('APPLOGGER', 'Core\Logger');
````

Are you proposing a change to the logger?

LuckyCyborg:
@daveismyname Nope. 

But you used to configure it in, well... configuration. Shall be at user pleasure to change or we hardcode into bootstrap?

daveismyname:
What I'm not clear about it how the logger becomes active why not use this:

````
set_exception_handler('Core\Logger::ExceptionHandler');
set_error_handler('Core\Logger::ErrorHandler');
````

LuckyCyborg:
@daveismyname The logger is loaded by **system/bootstrap.php**, and if you don't want it configurable, right that will be put into bootstrap

LuckyCyborg:
As in to use the app/Config.php only for config options, all the commands move to bootstrap (which also include that file)

daveismyname:
ah I understand now, the timezone should by Europe/London by default I know it will need changing for other timezones but it needs to start somewhere.

LuckyCyborg:
@daveismyname Sure, but I made a dump for my current config file. Into example will be London, of course

daveismyname:
I figured that would be the case, OK go ahead.

PR #735, opened 2016-04-15
Improve the Routing, Controllers and View rendering

LuckyCyborg:
This pull request introduce a new composite method into Core\View, capable to fetch the View rendering and able to work also with Modules. Its usage is simple:

```php
$content = View::fetch('Page/Show', $data, 'Pages');

echo $content;
```
OR
```php
$data['content'] = View::fetch('Welcome/SubPage', $data);

View::renderTemplate('default', $data);
```

To note that the third parameter of the View::fetch() is an optional Module name, permitting a style similar with View::renderTemplate(), without specifying the full path into Module.

Using this new method permit even to use an alternative rendering style, using Layouts.

In other hand, this pull-request introduce a simple and non-invasive post-processing support into Controllers execution, having a new Method, called 'after', which is automatically executed when the current Action return a value different of null or boolean.

This post processing ability can be very useful in the **RESTful** Controllers, for example doing:

```php
public function index()
{
    $data = array(
         'success' = true;
         ...
    );

    return $data;
}

public function show($id)
{
    $data = array(
         'success' = true;
         ...
    );

    return $data;
}

protected function after($data)
{
    header('Content-Type: application/json');

    echo json_encode($data);
}
```

Also, this post processing can be very useful when it is used a Layout style rendering, to not write again and again the same triplets; as in example:

```php
public function index()
{
    $data['title'] = $this->language->get('welcomeText');
    $data['welcomeMessage'] = $this->language->get('welcomeMessage');

    // Render the View and fetch the output in a data variable.
    $data['content'] = View::fetch('Welcome/Welcome', $data);

    return $data;
}

public function subPage()
{
    $data['title'] = $this->language->get('subpageText');
    $data['welcomeMessage'] = $this->language->get('subpageMessage');

    // Render the View and fetch the output in a data variable.
    $data['content'] = View::fetch('Welcome/SubPage', $data);

    return $data;
}

protected function after($data)
{
    View::renderTemplate('default', $data);
}
```

To note that the returned value of the current Action is passed to post-processing method as parameter and that this pull-request have **no impact** to current API of the framework.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-
+.idea
 # Framework specific
 vendor/
 composer.lock
--- a/system/Core/Controller.php
+++ b/system/Core/Controller.php
@@ -27,7 +27,17 @@ abstract class Controller
      */
     public function __construct()
     {
-        /** initialise the language object */
+        /** Initialise the Language object */
         $this->language = new Language();
+    }
+
+    /**
+     * Method automatically invoked after the current Action, when it not return a
+     * null or boolean value. This Method is supposed to be overriden for using it.
+     *
+     * Note that the Action's returned value is passed to this Method as parameter.
+     */
+    protected function after($data)
+    {
     }
 }
--- a/system/Core/Controller.php
+++ b/system/Core/Controller.php
@@ -37,7 +37,7 @@ abstract class Controller
      *
      * Note that the Action's returned value is passed to this Method as parameter.
      */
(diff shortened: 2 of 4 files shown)


LuckyCyborg:
@daveismyname Please edit the release notes, as there is a small error: the function **after** should be **public**, not **protected** into examples.

daveismyname:
thanks I've changed this.

PR #737, opened 2016-04-15
Make the third parameter of View::render() to be an (optional) Module name

LuckyCyborg:
#### Rationale

Right now, **View::render()** use to have as third parameter an (legacy?) array of errors, while that parameter is not really necesary now, while the View rendering extract the parameters from the second parameter $data. 

In fact, if **$data** contains an element called "error", it will override the array passed as third parameter, logic prone to generate mysterious rendering errors per-se.

In other hand, passing as third parameter the optional name of the Module, just like into **View::fetch()** is of nature to simplify the View class design (literally making obsolete the **View::renderModule()**), to offer a consistent API with **View::fetch()** and **View::renderTemplate()** and, why not? to offer an alternative to writing **View::renderTemplate()**'s long commands.

Then, instead of:
```php
View::renderModule('Pages/Views/Page/Index', $data);
```
we can use directly:
```php
View::render('Page/Index', $data, 'Pages');
```

Which is directly consistent with:
```php
$content = View::fetch('Page/Index', $data, 'Pages');
```
and similar to:
```php
View::renderTemplate('header', $data, 'Admin');
```

To note that pull request introduce a very small API incompatibility, about the **View::render()** third parameter, when the $error is passed this way, which can be resolved very simple (and in a more safe way) integrating the **$error** into **$data**, with:
```php
$data['error'] = $error;
```

### DIFF
--- a/system/Core/View.php
+++ b/system/Core/View.php
@@ -80,7 +80,7 @@ class View
      */
     public static function renderModule($path, $data = false, $error = false)
     {
-        if($error !== false) {
+        if(($error !== false) && isset($data['error'])) {
             // Adjust the errors data handling, injecting it into $data.
             $data['error'] = $error;
         }
--- a/system/Core/View.php
+++ b/system/Core/View.php
@@ -48,8 +48,6 @@ class View
      */
     public static function render($path, $data = false, $module = false)
     {
-        self::sendHeaders();
-
         // Pass data to check and store it.
         $data = self::prepareData($data);
 
@@ -66,6 +64,8 @@ class View
         }
 
         // Render the View.
+        self::sendHeaders();
+
         require APPDIR .$filePath;
     }
 
@@ -99,8 +99,6 @@ class View
      */
     public static function renderTemplate($path, $data = false, $custom = TEMPLATE)
     {
-        self::sendHeaders();
-
         // Pass data to check and store it.
         $data = self::prepareData($data);
 
@@ -111,6 +109,9 @@ class View
         foreach ($data as $name => $value) {
             ${$name} = $value;
         }
+
+        // Render the Template.
+        self::sendHeaders();
 
         require APPDIR .$filePath;
     }
--- a/system/Core/View.php
+++ b/system/Core/View.php
@@ -81,7 +81,7 @@ class View
     public static function renderModule($path, $data = false, $error = false)
     {
         if(($error !== false) && ! isset($data['error'])) {
-            // Adjust the errors data handling, injecting it into $data.
+            // Adjust the $error parameter handling, injecting it into $data.
             $data['error'] = $error;
         }
 
--- a/system/Core/View.php
+++ b/system/Core/View.php
@@ -87,7 +87,7 @@ class View
 
         if (preg_match('#^(.+)/Views/(.*)$#i', $path, $matches)) {
             // Render the Module's View using the standard 'render' way.
-            self::render($matches[1], $data, $matches[0]);
+            self::render($matches[2], $data, $matches[1]);
         }
     }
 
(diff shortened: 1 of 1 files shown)


jimgwhit:
@daveismyname Rather why not pass module as a optional fourth parameter leaving error alone.

LuckyCyborg:
@jimgwhite

>  Rather why not pass module as a optional fourth parameter leaving error alone.

There is my take: from reasons of API consistency and to close a subtle logic problem into View::render(). Example:

```php
$data['title'] = 'Hello to Jim';
$data['error'] = 'Hello, Jim!';

$error[] = 'This is the first error';
$error[] = 'This is the second error';

// Render the View
View::render('Welcome/Index', $data, $error);
```
The displayed error into page will be:
```
Hello, Jim!
```

In other hand, adding the Module name make no sense, considering that the third one is not really useful into context of Nova and prone for... errors. 

Also, I believe that View API will become more clear and consistent, like I said (with examples), into first post.

jimgwhit:
@daveismyname and @LuckyCyborg When a PR is done could it be possible to put a PR to the documentation as well for changes. And if a significant change perhaps an example.

LuckyCyborg:
@jimgwhit The only significant change is the third parameter of **View::render()** which now represent the Module Name (if any) instead or an array of errors, making **View::renderModule()** obsolete and even if I provided a wrapper for it, I believe that in future that method should be removed as being redundant.

In the mean time, the error array can by simply well integrated into $data, with no changes from Views code POV.

Kingorgg:
I agree with replacing the $error variable with the optional Module name.
Makes sense to just pass the errors through the $data array.

daveismyname:
I can see how this makes sense to keep everything consistent, this will break existing code, namely $error will have to be removed from render and errors will need to be added to a $data in order to pass them correctly.  

why have this:

````
$filePath = str_replace('/', DS, "Views/$path.php");
````

instead of 

````
$filePath = Views/$path.php;
````

since it's going to be used with APPDIR I don't see the benefit.

speaking of consistency should'nt renderTemplate also have $withHeaders

also the if statements need a space after the if but before the open parenthesis on lines 57, 69 and 85 to be psr compliant.

LuckyCyborg:
@daveismyname 

 > can see how this makes sense to keep everything consistent, this will break existing code, namely $error will have to be removed from render and errors will need to be added to a $data in order to pass them correctly.  

In other hand, consistent should be to avoid the possibility to have mysterious errors made by unwanted variables overriding. Let's be clear: you chose to have $data extraction. In one way or another, there should be only one (array). What you chose, without confusing your end-users? 
-------------------------------------
why have this:

````
$filePath = str_replace('/', DS, "Views/$path.php");
````

instead of 

````
$filePath = Views/$path.php;
````

-----------------------------------
Just a safe check that the DIRECTORY_SEPARATOR is properly used.

> since it's going to be used with APPDIR I don't see the benefit.

It has already a properly path, while the next part is user entered variable.

> speaking of consistency should'nt renderTemplate also have $withHeaders

Because make no sense to send headers a View fetch (where the output is already captured), and render(ing) is used by fetch too. Or you want to made it **fetch -> render** ? Fast to implement but a performance penalty. :smirk:

In other hand, renderTemplate shall send always the headers in the first hit. Or we should invent an new class "Core/Template" and to split the code, to be more clear?

> also the if statements need a space after the if but before the open parenthesis on lines 57, 69 and 85 to be psr compliant.

Feel free do adjust the PSR compliance but my really old eyes can't see good "!" while it is stuck to function or whatever.

daveismyname:
nope I don't see myself using fetch it should be there as an option only not a requirement.

PR #742, opened 2016-04-18
Update View.php

jimgwhit:
Moves the 
```
$data = self::prepareData($data);
```
in an if statement, that way if someone isn't using hooks, it is just better logic.
Like this
```
if ($data !== false) {
            $data = self::prepareData($data);
            foreach ($data as $name => $value) {
                ${$name} = $value;
            }
        }
```
In the render and renderTemplate methods.

### DIFF
--- a/system/Core/View.php
+++ b/system/Core/View.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * View - load template pages
  *
@@ -15,6 +16,7 @@ use Helpers\Hooks;
  */
 class View
 {
+
     /**
      * @var array Array of HTTP headers
      */
@@ -49,8 +51,13 @@ class View
      */
     public static function render($path, $data = false, $module = false, $withHeaders = true)
     {
-        // Pass data to check and store it.
-        $data = self::prepareData($data);
+        // Pass data to check and store it and Prepare the rendering variables..
+        if ($data !== false) {
+            $data = self::prepareData($data);
+            foreach ($data as $name => $value) {
+                ${$name} = $value;
+            }
+        }
 
         // Prepare the (relative) file path according with Module parameter presence.
         if ($module !== false) {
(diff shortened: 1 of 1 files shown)


jimgwhit:
I noticed the prepareData method uses $hooks = Hooks::get();  
And since Hooks are optional, I thought it would be good to bypass the method if not needed in certain applications. But all still work if Hooks are used.

LuckyCyborg:
@jimgwhit Here is my take:

One can have data returned via Hooks (i.e. custom  CSS), even if there is no $data passed to render(ing) Method.** That's why I not agree with you patch.**

In other hand, Yes, there is a issue with those hooks: they are executed every time a rendering method is called (read: walking the Modules and loading the associated files), even this is required a single call of them to collect the data. Then, they are executed three times under standard Nova rendering (or two, in the alternative Layout style), thing that can considerable slowdown the application when there are a huge number of modules.

So, maybe the actual Hooks integration is not the luckiest one...
 
A much better alternative is to invent a **View::share($key, $value)** which store information in a secondary shared array, merged to data at run time, and to load the Hooks in a pre-processing "before" method of the current Controller. The result will be the **calling the Hooks one time** for rendering, considerable speeding up the application, when a huge number of Modules is used.

Yeah, sometimes most times the simply things are not the best that you can have. :smirk:

jimgwhit:
@LuckyCyborg and @daveismyname yeah the method prepareData
```
    public static function prepareData($data)
    {
        $data = ($data !== false) ? $data : array();
        // Run the associated Hooks.
        $hooks = Hooks::get();
        $data['afterBody'] = $hooks->run('afterBody', $data['afterBody']);
        $data['css']       = $hooks->run('css', $data['css']);
        $data['js']        = $hooks->run('js', $data['js']);
        return $data;
    }
```
Is currently called weather someone is using hooks or not, I thought it would be a good idea to skip this call if hooks were not being used.  I also agree with @LuckyCyborg that the hooks could stand a revamp.  Still having it optional weather to use them or not.  But to have a call that is right in it:
```
$hooks = Hooks::get();
```
And if hooks aren't even being used in an application is my main concern.  Maybe this will work until you and @LuckyCyborg can somehow revamp the call for hooks.  Have a method that is called only if hooks are being used.  But this should bypass that if hooks are not being used.

LuckyCyborg:
@jimgwhit Nope, I will not agree with your pull request, even it point to a real optimization issue.

Please leave to me this problem, anyway you are doing it right.

LuckyCyborg:
@jimgwhit We agree to **close** this pull-request being incorrect and superseded by #743 and generating an non useful conflict eventually?

daveismyname:
Even if the hooks are not in use the data is returned.

PR #743, opened 2016-04-18
 Add support for pre-processing into Core\Controller and improve Core\View

LuckyCyborg:
#### Rationale

In the **Core\View** exists a (possible) performance issue, as the associated Hooks are executed every time when a rendering Method is called, then for a typically triplet rendering, they will be called three times, even the results are the same.

What is wrong with that? When in the application exists a big number of Modules, they will be processed in the loop and modules files loaded, Controllers are instantiated and methods executed, possibly slowing down considerable the rendering.

The proposed solution give a single shoot to the Views associated Hooks, into Controller, before executing the requested Method, and store the resulted data in a Views shared storage, from where is loaded and added for rendering execution. This style avoid multiple calling of those Hooks, as in for every rendering call.

In the Core\View is introduced a new Method, called "share", permitting to set data variables which are shared by any called rendering method. For example, it is possible to have:

```php
View::share('title', 'Shiny Title');
```
Making the variable 'title' available in any rendering command. This command is also very useful, as being used by design to setup the data coming from the Hooks execution.

In other hand, a generic pre-processing method called "before" and automatically executed before the requested action, is introduced and it by default execute the Views associated Hooks and share their result to **Core\View**. Yet, **it doesn't introduce API changes** and **it can be safely ignored** if there is no need to customize it.

Still, that pre-processing permit to fine tune the Controller behavior, for example hosting logic to check the user authentication or do whatever pre-processing, i.e. checking access codes, without requiring to add commands to every method, simplifying the code.  

To note that in a "before" method are available the requested method name and the given parameters, in read-only mode, via two simple additional getter methods. 

An complex example of its chained usage can be:

```php
// Into a Base Controller

protected function before()
{
    // Check if the User is authorized to use the requested Method.
    switch($this->method()) {
        case 'login':
        case 'forgot':
        case 'reset':
            break;
        default:
            if (Session::get('loggedIn') == false) {
                Url::redirect('login');
            }
    };

    // Leave to parent's Method the execution decisions.
    return parent::before(); // <-- there are loaded the Views associated Hooks.
}
```

```php
// Into an Admin Controller

protected function before()
{
    // Check the CSRF token for select Methods over POST, e.g: add/edit/delete.
    switch($this->method()) {
        case 'index':
        case 'show':
            break;
        default:
            if (Request::isPost() && ! Csrf::isTokenValid()) {
                Url::redirect('login');
            }
    };

    // Leave to parent's Method the execution decisions.
    return parent::before(); // <-- there is checked the User Authorization.
}
```

Another more complex example, for a Controller for private files serving; with access authorization and time-bomb (limit the access time):

```php

protected function before()
{
    // Only the authorized Users can arrive to the Methods from this Controller.
    if (Session::get('loggedIn') == false) {
        Url::redirect('login');
    }

    //  
    if($this->method() == 'files') {
        $params = $this->params();

        if(count($params) != 3) {
             header("HTTP/1.0 400 Bad Request");

             return false;
        }

        // The current timestamp.
        $timestamp = time();

        // File access validation elements.
        $validation = $params[0];
        $time       = $params[1];
        $filename   = $params[2];

        $clientip = $this->getIpAddress();

        $hash = hash('sha256', $clientip.$filename.$time.FILES_ACCESSKEY);

        if((($timestamp - $time) > FILES_VALIDITY) || ($validation != $hash)) {
            header("HTTP/1.0 403 Forbidden");

            return false;  
        }
    }  else if (Request::isPost()) {
         // All the POST requests should have a CSRF Token there.
         if (! Csrf::isTokenValid()) {
             Url::redirect('login');
         }
    }

    // Leave to parent's Method the execution decisions.
    return parent::before(); // <-- there are loaded the Views associated Hooks.
}
``` 

Also, the part of Method flow execution was moved into **Core\Controller::execute()**, presenting the big advantage to simplify the Routing and permitting to have the before/after methods **protected**, safe against their execution from outside.

Finally, into Core\Controller was introduced an new method called "trans", wrapping up the Language translation. Then, permitting commands like:

```php
// Actual style
$data['title'] = $this->language->get('welcomeText');
$data['welcomeMessage'] = $this->language->get('welcomeMessage');

// New style
$data['title'] = $this->trans('welcomeText');
$data['welcomeMessage'] = $this->trans('welcomeMessage');
```

Again, to note that **this pull request do NOT introduce any API break.**

#### Someone asked about an example of using "View::share()"

```php
public function index()
{
    $title = $this->language->get('welcomeText');

    $data['welcomeMessage'] = $this->language->get('welcomeMessage');

    // Make $title available in all Views rendering
    View::share('title', $title);

    View::renderTemplate('header'); // <-- No need for $data in this case
    View::render('Welcome/Welcome', $data);
    View::renderTemplate('footer'); // <-- No need for $data in this case
}
```

#### What is difference between applying pre-processing in "__constructor()" and "before()" ?

Comparative with running checks in Class Constructor, the **before()** method is executed WHEN the **requested Method** is know being **a valid one**, both as in existing and being "callable", and it have at its disposition the requested Method name and the associated Parameters; both passed to by Routing. Then, **before()** can accurate tune the Controller behavior, depending on requested Method.

#### Is required to implement the methods "before()" and "after()" in every Controller ?

No. You can safely completely ignore their existence if you don't want to fine tune the Controller's behavior via that Middleware.

#### What if I want a base Controller which do NOT call the Views associated Hooks ?

Just override the **before()** like bellow in a Base Controller, e.g. **App\Core\Controller**

```php
namespace App\Core;

use Core\Controller as BaseController

class Controller extends BaseController
{
    public function __construct()
    {
        parent::__construct();
    }

    protected function before()
    {
        return true;
    }
}
```

Then use this Class as base for your Controllers.


jimgwhit:
@LuckyCyborg  and @daveismyname But this line 
~~~
        $data = self::prepareData($data);
~~~
Is still in the render and rendertemplate methods, but it's not needed if someone isn't using hooks at all. That was my concern. I still think putting that line inside an if statement will alleviate that part.  
Why call prepareData if not needed?  I still think having this in an if clause alleviate that problem and adds to your solution as well for efficiency. 
~~~
if ($data !== false) {
            $data = self::prepareData($data);
            foreach ($data as $name => $value) {
                ${$name} = $value;
            }
        }
~~~
Otherwise that method is still called every time render or rendertemplate is used whether it is needed or not.

LuckyCyborg:
@jimgwhit And what is wrong with this line?

```php
$data = self::prepareData($data);
```

In the design given by this PR, there is no such call for Hooks into **prepareData()**, but it is required adjust the $data and to merge the current data with the shared one.

In fact, there is no call for Hooks in the entire **Core\View** class, right now, under this PR.

The Views associated Hooks run is move in a method called "before" in the Core\Controller and this call of Hooks is made exactly **one time**, before to execute the requested method.

That's all can be done under Dave's design of using Hooks for adding blocks, CSS and JS to Views.

In other hand look at the last statement of my PR description, were I detailed a very simple mode the eliminate completely the call of Views associated Hooks, probably exactly what you want.

LuckyCyborg:
@jimgwhit BTW, after this pull-request I have a design for **Routing Filters**, permitting things like:

```
Router::filter('auth', function($route) {
    if (Session::get('loggedIn') == false) {
        Url::redirect('login');
    }

     return true;
});
```

```php
Router::any('admin(/dashboard)', array(
    'before' => 'auth',
    'uses' => 'App\Controllers\Admin\Dashboard@index'
));
```

LuckyCyborg:
@daveismyname Seriosly? **The StyleCI analysis has failed - 66 files need addressing** ?!? he want to "fix" even the PhpMailer code? :smiling_imp: 

"Fix" you first the style of the Framework and I will love to "fix"  the style my pull requests...

And please disable the StyleCI for my branch: **3.0-nova**. I'm used as any programmer, for a well used style. I hope you understand.

jimgwhit:
@LuckyCyborg  this PR is for nova Framework master, correct to address the hook issue?

LuckyCyborg:
@jimgwhit This pull request touch also the triple/double Hooks calling by page rendering.

Its objective is to complete the middleware support for Core\Controller, via a new pre-pocessing method **before()** and optimizations into Routing.

After this PR, the next one, in waiting, will introduce the **Route Filters** and other improvements. 

As usual, I recommend you to ignore things which confuse you.

For example you can completely **ignore the existence** of the **before()** and **after()** methods into **Core\Controller**.

jimgwhit:
@LuckyCyborg I realize this PR takes care of the hooks problem the many extra calls. Any idea when
@daveismyname will push this on through?

LuckyCyborg:
@jimgwhit 

> I realize this PR takes care of the hooks problem the many extra calls. 

Yeah, you can put the things also this way. As consequence of this pull request, who introduce a better data management into Core\View and moving the Views associated Hooks calling to Middleware, the issue of multiple Hooks calling is resolved, as better can be. 

Also introduce a simple method to eliminate complete that Hooks calling, like I demonstrated in a example.  

> Any idea when @daveismyname will push this on through? 

I wonder me too... :smirk: 

After all, this pull-request do not introduce API breaks, being much more about optimization and new features. Also I wait for its merging to start porting of the **Router Filters** to Nova's Routing.

daveismyname:
Apologies for the delay I have a lot going on right now.

PR #772, opened 2016-04-23
Introducing Core\Template, a View specialized to handle Template files

LuckyCyborg:
### Rationale

In almost any of Big Houses of frameworks, the **View** Class deal with a single **VIEWS PATH**. 

Under Nova, behind **Views**, we have also the concept of **Templates**, which have their own design and concept. 

Dealing with Template files in the same **View** class make us to complicate its design an can be confusing for end-users some of them looking to have problems to understand the inherited difference between a **View file** associated to a Controller and its Method, and a Template file, who give you (parts of) Layouts, even when using the ol'good **Triplet Style Rendering**.  

I consider that is not better to continue this confusion maker situation even into the New Style Rendering API and it better to have a dedicated class who deal with the Template Files.

Enter the **Core\Template**, a View specialized to handle Template files; basically identical with (its parent) View, but with its own **make()** builder and no Compatibility Layer, being destined to be used New Style API only.

### Usage

The usage of **Core\Template** is very simple. Instead to execute:
```php
$template = View::makeTemplate('default', $data, 'Admin');
```
You can simply do:
```php
$template = Template::make('default', $data, 'Admin');
```

From there, you can process it as usual, but the code will be much more clear and non confusional, **Template** dealing with the files from **app/Templates**, while the **View** deal **only** with the files from **app/Views**, and Modules based Views.

### Notes

Before to merge this pull request, I have two questions for you, @daveismyname .

**1** Is there a need for real to introduce into Compatibility Layer the support for **View::makeTemplate()**, previously introduced into last merged pull request? 

**2** It there a need for real to continue to support into Compatibility Layer the method **View::renderModule()**, (which complicate the code per-sè), considering that since **v3.6.0** the method **View::render()** have a native support for Modules based Views ?

The changes are trivial to port the existent code to **View::render()** modular rendering, as in specifying the third parameter, instead of doing:
```php
View::renderModule('Blog/Views/Blogs/Index', $data);
```
Yo have to change to:
```php
View::render('Blogs/Index', $data, 'Blog');
```

If you agree to not introduce into Compatibility Layer the **View::makeTemplate()** and to remove support for **View::renderModule()**, I will add another patch, capable to simplify both **Core\View** and **Core\Template**.

### DIFF
--- a/system/Core/Controller.php
+++ b/system/Core/Controller.php
@@ -10,6 +10,7 @@ namespace Core;
 
 use Core\Language;
 use Core\View;
+use Core\Template;
 use Helpers\Hooks;
 
 /**
@@ -126,9 +127,9 @@ abstract class Controller
         //
         // Execute the default Template-based rendering of the given View instance.
 
-        if (! $data->isTemplate() && ($this->layout !== false)) {
-            // The View instance is NOT a Template, but have a Layout is specified.
-            View::makeTemplate($this->layout, array(), $this->template)
+        if ((! $data instanceof Template) && ($this->layout !== false)) {
+            // The View instance is NOT a Template, but we have a Layout specified.
+            Template::make($this->layout, array(), $this->template)
                 ->withContent($data)
                 ->display();
         } else {
--- a/system/Core/Template.php
+++ b/system/Core/Template.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * Template - a View specialized for handling the Template files.
+ *
+ * @author Virgil-Adrian Teaca - virgil@giulianaeassociati.com
+ * @version 3.0
+ */
+
+namespace Core;
+
+use Core\View;
+
+
+class Template extends View
+{
+    /**
+     * Constructor
+     * @param mixed $path
+     * @param array $data
+     *
+     * @throws \UnexpectedValueException
+     */
+    public function __construct($path, array $data = array())
+    {
+        parent::__construct($path, $data);
+    }
+
+    /**
+     * Create a Template instance
+     *
+     * @param string $path
+     * @param array $data
+     * @param string $custom
+     * @return Template
+     */
+    public static function make($path, array $data = array(), $custom = TEMPLATE)
+    {
+        // Prepare the file path.
+        $filePath = str_replace('/', DS, "Templates/$custom/$path.php");
+
+        return new Template(APPDIR .$filePath, $data);
+    }
+
+    /**
+     * Magic Method for handling dynamic functions.
+     *
+     * This method handles calls to dynamic with helpers.
+     */
+    public static function __callStatic($method, $params)
+    {
+        // No Compatibility Layer exists there; nothing to do.
+    }
+
+    /**
+     * Compat Layer - Render a Module View file.
+     *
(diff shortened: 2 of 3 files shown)


daveismyname:
yes there is a need all users will be using renderTemplate to loose it will break those installs, the changes in your last pr supported old style and new the same should apply to this.

Not to mention it makes all videos I've done will be advocating processes that no longer work.

One of the reasons I haven't recorded any videos recently is the fastly paced changes makes it hard to keep videos and docs up to date.

I've got the new changes from the last pr in my install although I'm still using the old style as they work perfectly for my needs so I don't need to change them there is no benefit for my use case.

The same will apply to this I will carry on using renderTemplate, updating them when the end result is going to be the same just seems like extra work for no real gain.

the renderModule could be marked as depreciated to alert uses to use render instead.

LuckyCyborg:
@daveismyname There is no loose of **View::renderTemplate()** and breaking the installs. 

The View's **Facade** will continue to support **View::renderTemplate()** as **usual**, no break included.

The **Core\Template** class is specific form the New Style API, and offering a more clear concept for **Views != Templates**.

Again, the use of the "old" style rendering **will not be influenced** by the presence of **Core\Template**, even it will be used internally to do the jobs.

daveismyname:
OK cool 👍

LuckyCyborg:
Then, the questions are:

1 . supporting the New Style's **View::makeTemplate()** , introduced into latest **v3.9.0** into Compatibility Layer or just nuke it.

2. How we deprecate **View::renderModule()** ? Replacing its functionality with displaying a notice with no functional work or keeping it functional as doing its job?

daveismyname:
as that's brand new I'd say replace it.

I'm not sure what the correct course of action is, have the display notice advising to use render instead?

LuckyCyborg:
@daveismyname How about adding to:

```php
echo '<p>Please use <b>View::render()</b> instead of <b>View::renderTemplate</b></p>';
```

daveismyname:
do you mean:

```php
echo '<p>Please use <b>View::render()</b> instead of <b>View::renderModule</b></p>';
```

Probably make sense to add this will be removed in a future version

LuckyCyborg:
@daveismyname Okay, I will add a patch to update the things...

LuckyCyborg:
@daveismyname Added. please look to it.

geomorillo:
So you just did nuke it... :+1:

LuckyCyborg:
@geomorillo First of all, I have the accept of @daveismyname , secondly I try to keep the View's Facade simple.

In other hand, I mussed that will confusing to have a **View::makeTemplate()** which return a **Template** instance.

Also, I loved to have the ability to do, for a complete custom rendering:
```php
Template::make('custom', $data, 'Admin')
    ->nest('content', 'Admin/Dashboard/Index', $data)
    ->display();
```
Finally, @daveismyname  the View/Template engines are just like.. I like. I will touch further only for eventually bugfixes.

Next on will be enhancing the **Heplers\Response** and in future we will have ability to do in a Route Filter:
```php
return Response::redirectTo('login')->withErrors('You are not authorized to access this page');
```

PR #781, opened 2016-05-05
Remove the Core\View@renderModule support, improve the logic on Core\View@offsetGet and the properties getter

LuckyCyborg:
This Pull Request introduce a new class, called **Core\BaseView**, which is the base for **Core\View** and **Core\Template**, and contains their common logic. This way, the code of Rendering is simplified.

To note that this Pull Request remove the Compat Layer for **View::renderModule()**, its functionality being supported completely by **View::render()**, as its third parameter being the optional **Module** name, and for now we throw a meaningful Exception for.

Also to note that removing the Compat Layer for **View::renderModule()** sensible simplify the **Core\View** and **Core\Template**. 

Finally, this Pull Request introduce improvements on Routing.

### DIFF
--- a/app/Controllers/Demo.php
+++ b/app/Controllers/Demo.php
@@ -35,8 +35,7 @@ class Demo extends Controller
             'param3' => $param3,
             'param4' => $param4
         );
-        echo '<pre>';
-        print_r($params);
-        echo '</pre>';
+
+        echo '<pre>' .var_export($params, true) .'</pre>';
     }
 }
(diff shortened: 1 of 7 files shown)


geomorillo:
go ahead no problem here

LuckyCyborg:
@geomorillo 

> go ahead no problem here

Thanks, but the only "shocking news" introduced by this pull request is removing the support for deprecated **View::renderModule()**

Everything else are optimizations.

daveismyname:
This change does not alter the normal use of render and renderTemplate does it?

LuckyCyborg:
@daveismyname Of course not. The following commands works as usual via View's Facade:

View::fetch()
View::render()
View::renderTemplate()

PR #786, opened 2016-05-08
Implementation of dynamic Response and Redirection management

LuckyCyborg:
This Pull Request introduce an improved **Response** API, in a Laravel-esque style, able to simplify the Framwork's Response management. Practically, is now possible to do in a Controller Method:

```php
// Create a Response instance with string content
return Response::make(json_encode($user));

// Create a Response instance with a given status
return Response::make('Not Found', 404);

// Create a Response with some custom headers
return Response::make(json_encode($user), 200, array('header' => 'value'));

// Create a response instance with JSON
return Response::json($data, 200, array('header' => 'value'));

// Create a 404 response with data (will be directly obtained a shiny themed Error Page)
return Response::error('404', array('message' => 'Not Found'));
```
It is also possible to use those Response instance in the **Route Filters**, case when the given Response will be send to browser and the further processing will be skipped. 

To note that **Response** class was moved to **Core** considering as being now a vital part of Framework, for its properly work, and not an (optional) Helper.

Also, there is introduced a new class, called **Core\Redirect**, built on top of **Core\Response**, which permit in Route Filters or Controller Methods commands like:

```php
// Create a Redirect Response to a location within the application
return Redirect::to('user/profile');

// Create a Redirect Response with a 301 status code
return Redirect::to('user/profile', 301);

// Create a Redirect Response and flash to the Session
return Redirect::to('profile')->with('message', 'Welcome Back!');
```

Someone can ask: what is wrong with **Url::redirect()** to need to return those **Redirect** instances from the Controller Methods and Route Filters?

The response is simple: while with a simple and clear API, yet powerful, the Routing and/or Controllers execution on Nova Framework is a complex thing, including even two stacked Middlewares.

Suddenly stopping the **Framework Execution Flow** in a middle of some Controller Method or Route Filter, is **not** the luckiest idea in the town, in the Nova's context. And exactly that **Url::redirect()** do, sending a redirection HTTP Header and unceremonious quitting the current execution.

In other hand, the new Response API permit further simplifications on Routing. For example, the command **Response::error()** is quite capable to display a themed/templated Error Page, without the help of an additional "Error Controller" or other "callbacks".


jimgwhit:
@LuckyCyborg you said  
> To note that Response class was moved to Core considering as being now a vital part of Framework, for its properly work, and not an (optional) Helper.

Does this PR break existing code for example will I have to change
```
Url::redirect() //style
```
to
```
return Redirect::to() //style
```
I don't understand why a return, normally a function returns something.  Where as Url::redirect() was direct to the point and just did it.

LuckyCyborg:
@jimgwhit 

> Does this PR break existing code

You are kidding, right? The **Url::redirect()** is an authoritative command, which send a redirection HTTP Header and literally quit the application execution in the middle of nowhere. Nothing in the Known Worlds can break that.

> I don't understand why a return, normally a function returns something. Where as Url::redirect() was direct to the point and just did it.

You read the pull-request, right? **Redirect::to()** returns a **Core\Redirect** instance, which is a glorified **Core\Response**. Like I said in description, when a Controller Method or Route Filter return a **Response** instance, in will be processed and graciously stopped the processing. 

That's much better than an authoritative quitting command, from reasons described. While not mandatory, I strongly recommend this way.

In other hand, **Core\Response** offer other shiny things, i.e. JSON encapsulation, etc.

jimgwhit:
@LuckyCyborg said  
> You read the pull-request, right? Redirect::to() returns a Core\Redirect instance  

But that is not your example, your example has
~~~
return Redirect::to() 
~~~
So could you give an example in a controller of where and how to use this with return statement rather than Direct?  
Again to me a return statement is used to return something from a method or at least a function inside of a method.

LuckyCyborg:
@jimgwhit

I will give you an example of CSRF Filter:

```php
Route::filter('csrf', function($route) {
    if (! Csrf::isTokenValid()) {
        // When CSRF Token is invalid, respond with Error 400 (Bad Request) 
        return Response::error(400);
    }
});
Route::filter('auth', function($route) {
    if (Session::get('loggedIn') == false) {
         // When User is not logged in, respond with a Redirect with code 401 (Unauthorized) 
        return Redirect::to('login', 401);
    }
});
```

> Again to me a return statement is used to return something from a method or at least a function inside of a method.

You return instances of **Core\Response** and **Core\Redirect**, which the Framework Middlewares and Execution Flow are able to interpret and send.

LuckyCyborg:
@jimgwhit

I will give you an example of CSRF Filter:

```php
Route::filter('csrf', function($route) {
    if (! Csrf::isTokenValid()) {
        // When CSRF Token is invalid, respond with Error 400 (Bad Request) 
        return Response::error(400);
    }
});
Route::filter('auth', function($route) {
    if (Session::get('loggedIn') == false) {
         // When User is not logged in, respond with a Redirect with code 401 (Unauthorized) 
        return Redirect::to('login', 401);
    }
});
```

> Again to me a return statement is used to return something from a method or at least a function inside of a method.

Practically, you return instances of **Core\Response** and **Core\Redirect**, which the Framework Middlewares and Execution Flow are able to interpret and to send to browser. The Controller's Methods and Route Filters Closures are functions like anyone, which can return whatever. And their return values are interpreted by the Framework.

jimgwhit:
@LuckyCyborg you have  return Redirect::to('admin/dashboard', 401);  
Would this work
~~~
 return Redirect::to('admin/dashboard');
~~~

LuckyCyborg:
@jimgwhit Yes, will send the (default) HTTP code **302** (Found)

PR #789, opened 2016-05-10
An improved Database API

LuckyCyborg:
This pull request introduce an improved Database API, in a Laravel-esque style, including a **QueryBuilder** and a simple but powerful **Model**, everything living in the namespace **'\Database\'** from isolation reasons.

To note that this new Database API doesn't replace any of existent classes, the actual **Core\Model** and **Helpers\Database** remaining untouched. Then an end-user can choose which Database API use in his application, with the only condition to not use both of them simultaneous, thing which will duplicate the Database connections.

Then new API is structured on three levels.

First one of the levels, being a Database **Connection**, built on top of **PDO**, which is available both via getting an instance:
```php
use Database\Connection;

$db = Connnection::getInstance();

$prefix = $db->getTablePrefix();

$data = $db->select("SELECT * FROM {$prefix}users");
```
and via a **Facade** which permit commands like:
```php
use DB;

$prefix = DB::getTablePrefix();

$data = DB::select("SELECT * FROM {$prefix}users");
```
Commands for insert/update/delete are available.

To note that the **Database\Connection** fetch the data into **objects** or **array of objects**.

The second level is represented by a very simple but powerful **QueryBuilder**, which permits commands like:
```php
use DB;

$users = DB::table('users')->get();

$users = DB::table('users')->where('role', '=', 'admin')->get();
```

The third level is represented by a simple, yet powerful, **Model**, which use the **Database\Connection** instead of **Helpers\Database**. It is recommended way to use the new Database API, and can have a structure like:
```php
namespace App\Models;

use Database\Model;

class Users extends Model
{
    protected $table = 'users';

    protected $primaryKey = 'userId';

    public function __construct()
    {
        parent::__construct();
    }
}
```
To note the two protected variables, which specify the Table name and Primary Key, which are mandatory to be configured, specially the Table name, the second one defaulting to **'id'**. 

A **Database\Model** is similar as usage with the **Core\Model**, but have the ability to transparently use the QueryBuilder methods, then permitting command like this:
```php
use App\Models\Users;

$model = new Users();

$users = $model->where('role', '=', 'admin')->get();
```

Also, the **Database\Model** offer associated **QueryBuilder** instances, as following:
```php
use App\Models\Users;

$model = new Users();

$query = $model->newQuery();

$users = $query->where('role', '=', 'admin')->get();
```


Kingorgg:
Looks good, can't wait to play around with it :+1:

LuckyCyborg:
@Kingorgg 

> Looks good, can't wait to play around with it :+1:

Thanks! But to remember that you have to choose which API you use overall application. If you choose this Database API the another one is a big no-no. :smirk:

Kingorgg:
I'll try it out in a different test project, to see which one I like most :smile:

LuckyCyborg:
He he! This on will give you a strong Laravel taste.

Think something inspired by Illuminate\Database, but with no ORM and bloody simplified... :smirk:

daveismyname:
this looks great I'm also going to try this out, do you have an example of using multiple tables / doing a join?

LuckyCyborg:
Right now, no. But fully supports JOIN and the API is similar with somewhat old Iluminate Database.

daveismyname:
OK thanks.

LuckyCyborg:
No problem.

PR #792, opened 2016-05-10
Custom js loading via the footer & Controller method

aidanayala:
Small adjustment to how the footer loads JS files.

Allows for page / method specific JS files to be defined in the controller & dynamically loaded by the footer.

A slightly more efficient way of loading JS that is only needed on certain pages.


LuckyCyborg:
Nope.

PR #793, opened 2016-05-11
Improves Csrf security using various functions

geomorillo:


Kingorgg:
Would it not be best to have:

````
    public static function genRandomNumber($size = 32) { // More code ... }
```

Instead of 

````
    public static function genRandomNumber() {
        $size = 32;
    }
```

?

daveismyname:
yes that would be more flexible.

PR #797, opened 2016-05-11
Implementation of a simple Auth system

LuckyCyborg:
This pull-request introduce a simple but efficient **Auth** system, in a Laravel-esque style.

All the classes of that **Auth** system live in the namespace **\Auth\** and is implemented also a reference structure for User Authentication in the **\App\** namespace.

To note that additional **Route Filters** are added, to support this reference implementation and the properly configuration of **Helpers\Encrypter**, as in valid **ENCRYPT_KEY**, is required.

Being an **Users Management**, of course, a Database is required and in **scripts/nova_users.sql** you can find the associated MySQL dump for the **users** Table. 

Important: as Database API is used the new one, living in the namespace **\Database\**, then if you choose to use this Auth System, you should use that new Database API in the whole application and to not touch the **Helpers\Database** instances.

The reference implementation contains a class called **App\Controllers\Auth** which control the Users **login** and **logout**, as a base for your Users management.

The main **Auth System** classes are:

**Auth\Auth** - aliased as **\Auth**, which contains a simple manager of the current **Guard** instance. 
**Auth\Guard** - the Authentication Guard, which handle the User authentication.
**Auth\Model** - an Users base **Model**, built on top of **Database\Model** and configurable via a Config entry.

The offered services are:
**User Login** with **remembering** - the solution use **a strong encrypted Cookie** for remembering the authenticated user. I.e.
```php
// Prepare the Authentication credentials.
$credentials = array(
    'username' => Request::post('username'),
    'password' => Request::post('password')
);

// Prepare the 'remember' parameter.
if (Request::post('remember') == 'on') {
    $remember = true;
} else {
    $remember = false;
}

if(Auth::attempt($credentials, $remember)) {
    // User is authenticated there.
}
```

**User Logout** - using a simple command:
```php
Auth::logout();
```

**Authentication check**, via a boolean command:
```php
if(Auth::check()) {
    // User is authenticated there.
}
```
**Current User information** - the Auth have the ability to retrieve the authenticated User information, in a **stdClass** instance, fetched via **PDO::FETCH_OBJ**. No ORM Entities or data arrays are supported. i.e.
```php
if(Auth::check()) {
    // User is authenticated there, retrieve its Record.
    $user = Auth::user();

   // Do something cool there.
}
```

Finally, to note that this Auth System do **NOT** offer support for **User Registration** or **Credentials Recovery**, being all about a simple, yet secure and strong, **User Authentication**.

### DIFF
--- a/app/Config.example.php
+++ b/app/Config.example.php
@@ -106,6 +106,7 @@ Config::set('database', array(
         'database'  => DB_NAME,
         'username'  => DB_USER,
         'password'  => DB_PASS,
+        'prefix'    => PREFIX,
         'charset'   => 'utf8',
         'collation' => 'utf8_general_ci',
     ),
@@ -114,7 +115,7 @@ Config::set('database', array(
 /**
  * Setup the (class) Aliases configuration.
  */
-Config::set('class_aliases', array(
+Config::set('classAliases', array(
     'Errors'        => '\Core\Error',
     'Language'      => '\Core\Language',
     'Response'      => '\Core\Response',
@@ -143,4 +144,15 @@ Config::set('class_aliases', array(
     'Tags'          => '\Helpers\Tags',
     'Url'           => '\Helpers\Url',
     'DB'            => '\Database\Facade',
+    'Auth'          => '\Auth\Auth',
+));
+
+/**
+ * Setup the Auth configuration.
+ */
+Config::set('authentication', array(
+    'model'      => 'Auth\Model',
+    // The used Database Table information.
+    'table'      => 'users'
+    'primaryKey' => 'id'
 ));
--- a/app/Config.example.php
+++ b/app/Config.example.php
@@ -92,7 +92,7 @@ define('SESSION_PREFIX', 'nova_');
 define('SITETITLE', 'Nova V3.0');
 
 /**
- * Define a 32 bits Encription Key.
+ * Define a 32 bits Encryption Key.
  */
 define('ENCRYPT_KEY', '');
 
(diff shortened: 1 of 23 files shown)


app/Config.example.php
```diff
@@ -92,7 +92,7 @@
 define('SITETITLE', 'Nova V3.0');
 
 /**
- * Define a 32 bits key
+ * Define a 32 bits Encription Key.
```
tomvlk: Typo

LuckyCyborg: Thanks!

system/Auth/Guard.php
```diff
@@ -0,0 +1,359 @@
+<?php
+/**
+ * Giuard - A simple Authentication Guard.
```
Kingorgg: Giuard -> Guard

LuckyCyborg: Thanks, will be fixed in next PR

geomorillo:
Nice nice

jimgwhit:
@LuckyCyborg can I get user info like  in laravel doc?
~~~

Accessing The Logged In User

Once a user is authenticated, you may access the User model / record:

$email = Auth::user()->email;
To retrieve the authenticated user's ID, you may use the id method:

$id = Auth::id();
To simply log a user into the application by their ID, use the loginUsingId method:

Auth::loginUsingId(1);
~~~

LuckyCyborg:
@jimgwhit I choose to implement only the basic commands to, and there is no ORM, then what returns Auth::user() is an **stdClass** instance fetched via **PDO::FETCH_OBJ** from database and which contains all the User record information. Then, that will work:

```php
$email = Auth::user()->email;
```

But, no **Auth::loginUsingId()** for you. If there is a need for, could be implemented in future.

PR #798, opened 2016-05-11
Changed: fa fa-check to glyphicon glyphicon-ok in Sessions Helper

Kingorgg:
Nova doesn't come with font awesome out of the box, so changed to bootstrap's check mark.


LuckyCyborg:
How about to include Font Awesome by default, instead?

It doesn't hurt... :smirk:

PR #803, opened 2016-05-11
Simplify the structure of the Reference Implementation of Users Authentication on Application

LuckyCyborg:
Like the title say.


jimgwhit:
Will we get to a point where there is no pull request for a while couldn't it be a stable version for at least 3 months? Is this turning into laravel again? People if you really want all these features just go to the Laravel site  and download laravel. I am scared to see what's coming next. I truly thought a framework was for people who knew how to implement their own Authentication. @LuckyCyborg why don't you just turn this into a generated code program like Xlinesoft has PHP Runner, it seems like that's the kind of audience that is  being drawn in people who wants everything and I mean everything done for them including all the code written. Some would actually be better off just hiring a programmer to do all their work for them may as well have the house cleaned by a maid. Get someone to mow your lawn to.  
I showed you several issues back where I implemented my own Authentication I mean for crying out loud can't people put together a little program with sessions and Implement some basic authentication on their own. What happened to keeping it simple and let people add in what they want as needed?

geomorillo:
Sorry to say this but you are exaggerating a little if you don't want to use it then implement the Auth system yourself, nobody is forcing you to use all the functionalities made, if you dont like the ones included then do it yourself, discard the others, also all backwards api compatibility is guaranteed with the facades so no break ups, The problem seems to be that you lack a little knowledge in oop, so this scares you, but thats how mvc frameworks are, they require advanced understanding of Mvc and OOP.
The only think i agree with you is that, it was maybe all too fast but i think the changes are good, you can use them or not is your decision.

jimgwhit:
@geomorillo oop has absolutely nothing to do with writing your own Authentication. Many moons ago before I even touched PHP I was writing Java beans servlet and JSP code at a medium sized trucking company. I don't like junk I like writing my own code.  
If you go back and do some research you will find out that there was actually errors in Laravel 4.2 Authentication. Now with the Frameworks Authentication being so complex in the background how would the average framework user who is actually trying to write code and not spend time finding errors in a framework supposed to know this.  
A framework is supposed to make things easier but when you cannot trust procedures and methods in a framework to be correct well you get the point that's why I write my own Authentication. A point is this framework was designed for people to  write their own stuff like Authentication.  
Are you telling me that Dave uses this framework and never implemented any kind of Authentication?  
I doubt that seriously I bet he has his own custom Authentication he uses. As do I.

LuckyCyborg:
@jimgwhit 

> Then how come the big supernova got shot down?

Your Big Mouth contributed to?

I would like to thanks you for that! .smirk:

tomvlk:
:smirk:*

LuckyCyborg:
@tomvlk 

> It's not only just an additional part, but at the same time an example for the framework, so :+1:

Surely, can be considered as practical example about how to be used the Framework. :smirk:

PR #806, opened 2016-05-12
Improve the Auth System

LuckyCyborg:
This pull request introduce a new **Auth\Guard** method, which check the User state as **guest** with a very simple usage:
```php
if(Auth::guest()) {
    // Do something, the User is not logged in.
}
```

Also, there are introduced optimizations of **Auth System** and a fix into **App\Controllers\Users@profile**.


geomorillo:
check my pr

LuckyCyborg:
@geomorillo Please close yours PR, I integrated there the change in checks order on **App\Controllers\Users@profile**

PR #807, opened 2016-05-12
first check if both passwords are equal

geomorillo:
i think this should be better also using the $password variable


LuckyCyborg:
You make a mistake: the $password contains the New Password

geomorillo:
yes you are right

LuckyCyborg:
@geomorillo Please, close this PR, I adapted the order in my PR

PR #826, opened 2016-05-15
Improve the Modules Configuration management

LuckyCyborg:
This pull-request introduce a simple but enhanced Modules configuration, using  a new Class called **Core\Modules**. Basically, now every Module can be considered a **min-App**, sporting its own **Config.php**, **Filters.php** and **Routes.php**, with functionality similar with the main ones.

To properly work, a App Configuration entry was added, and there should be specified the **active** Modules, the **Core\Modules** looking only for those specified by name there. E.g.

```php
Config::set('modules', array(
    'Users',
    'Blog',
    'Articles',
));
```

### DIFF
--- a/app/Config.example.php
+++ b/app/Config.example.php
@@ -108,6 +108,13 @@ define('ENCRYPT_KEY', '');
 define('CACHEPATH', APPDIR .'Cache');
 
 /**
+ * Setup the Active Modules
+ */
+Config::set('modules', array(
+    //'Users',
+));
+
+/**
  * Setup the Class Aliases configuration.
  */
 Config::set('classAliases', array(
--- a/system/Core/Modules.php
+++ b/system/Core/Modules.php
@@ -14,7 +14,7 @@ use Core\Config;
 class Modules
 {
     /**
-     * There we load the Modules configuration.
+     * The loaded Modules configuration.
      *
      * @var array
      */
--- a/system/Core/Modules.php
+++ b/system/Core/Modules.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * Modules Manager - class responsible to Modules management.
+ *
+ * @author Virgil-Adrian Teaca - virgil@giulianaeassociati.com
+ * @version 3.0
+ */
+
+namespace Core;
+
+use Core\Config;
+
+
+class Modules
+{
+    /**
+     * There we load the Modules configuration.
+     *
+     * @var array
+     */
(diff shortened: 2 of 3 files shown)


southcoastweb:
Morning,
I use modules as i find them easy to move from site to site and they work.
Ive just updated to latest version on my testing nova site and updated the config file to include the config::set(modules.... but i have nt added anything to this.
My modules are still working perfectly fine. Do i need the config bit? or should i remove the <name>.module.php file which currently controls the routes?
Im not 100% certain how to implement the changes (if i need to) in this PR ?

LuckyCyborg:
The **app/Config.php** entry is necessary to exists, even as empty array.

Yet, on the Modules specified in this Config array, the Framework will load, if exists, a series of files, called **Config.php, Filters.php and Routes.php**, per every Module, with configuration roles similar to those existent on App based directory.

In those files you can specify additional Config options, Route Filters or Routes, per Module, without the need to call some Hooks for that. Then, supposedly, much faster to load and more versatile.

Think about a Module as a mini-App, with its own complete Configuration, similar with the main one.

southcoastweb:
So this is an improvement that will better the way modules work/ can work without breaking the current way. 👍 
Maybe when Dave gets a chance or if you get a moment a full demo would be great with all the features working.   
When modules were introduced into 2.2 it was a great improvement for us.
We tend to write most sites now 'modulated' with SMVC for ease of upgrading.
Im just going through the process of doing our first 'production site' with Nova and rewriting most of our modules.

LuckyCyborg:
If you are going to rewrite your Modules, I would like to raise your attention about the Nova's **Events** dispatching, capable to send and receive (arbitrary) Data.

That, and the new Configuration support, will simplify much your Modules and their inter-communication.

southcoastweb:
Sounds good. 
Are there any guides yet to doing that ? 
I will try to baffle my way through it otherwise.

LuckyCyborg:
The Event dispatching is similar with those existent on **Laravel 4.x**, then in general that documentation apply, and after all, their usage is triviall simple. 

But, of course, you can also look on our pull requests and release notes.

southcoastweb:
ok great. I will pop over to laravel to get the documentation and your release notes

southcoastweb:
Hi @LuckyCyborg 
I ve just seen the pull request to move the users to modules..
Perfect for me to follow along.
U r a 🌟 
Thanx

PR #827, opened 2016-05-16
Base security updated

geomorillo:
-Improved version of st_random(), basically added a better fallback function if no openssl found
-Used Encrypter:randomBytes() in Csrf
-Removed mv_crypt from Encrypted
-Created fallback genRandomBytes for other functions


LuckyCyborg:
I would like about **'assets/image.png'**

When executed via **public** as document root, this path does not exists.

Should be an absolute path:

```
ROOTDIR .'assets/image.png'
```

geomorillo:
thanks

LuckyCyborg:
@geomorillo Just wondering: WHY we need an Entropy generation?

If the named Hosting Service lack both OpenSSL and MCrypt support, its about some trash payed next to nothing or even gratis. 

We have no duty to make the Framework to run even in a Fridge, right?

geomorillo:
Then why the fall back in your function? 
return substr(str_replace(array('/', '+', '='), '', base64_encode($bytes)), 0, $length); 
i can say the same about this code...
The answer to my question is the answer to your question.

geomorillo:
Then why the fallback in your function
$pool = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
    return substr(str_shuffle(str_repeat($pool, 5)), 0, $length);
I should ask the same...
why bother checking is openssl is present at all?
if (function_exists('openssl_random_pseudo_bytes')) {

LuckyCyborg:
I suppose that there should be OpenSSL, then, as last case, in some testing environment, to produce some random code.

geomorillo:
that is my answer

LuckyCyborg:
> that is my answer

Looks complicated and overkill. Like hunting mouses with guns.

geomorillo:
Well if you consider not including i dont mind..

LuckyCyborg:
Idea is good, but I consider the own entropy generator overkill. A fallback to mcrypt is more that enough, in my opinion.

tomvlk:
Please stick to the OpenSSL and fallback to mcrypt and no more.

geomorillo:
then i should change it again..

LuckyCyborg:
If you are kind... :smirk:

PR #841, opened 2016-05-18
Template style alignment and Update Persian language

amiroperator:
Implementation of a simple direction system for template style
Update Persian Language


LuckyCyborg:
Instead of
```php
$this->dir    = $info['dir'];
```
You should have 
```php
$this->dir    = $info['direction'];
```
According with you changes on Config.php

But I recommend also to use 'direction' instead of 'dir' into Language class

LuckyCyborg:
Please format **Language::direction()** as following:
```php
public static function direction($code = LANGUAGE_CODE)
{
    $code = self::getCurrentLanguage($code);

    $languages = Config::get('languages');
        
    if(isset($languages[$code]) && ! empty($languages[$code])) {
         $info = $languages[$code];

         return $info['dir'];
    }

    return 'ltr';
}
```
Is more readable.

amiroperator:
I tested that.
this code doesn't work because self::getCurrentLanguage($code); returns a first letter uppercase like Fa

LuckyCyborg:
Now does NOT.

pandory-network:
@LuckyCyborg 

`if (preg_match ('/[a-z]/', $cookie) && in_array($cookie, array_keys($languages))) {
     Session::set('language', ucfirst($cookie));
}`

set language cookie ucfirst

LuckyCyborg:
Also, I tested a very simple thing:

Changing : 

```
<html lang="en">
```
TO
```
<html lang="en" dir="ltr">
```
and guess what? The page content switched to right alignment.

To configure automatically, we can do like bellow
```php
<html lang="<?php echo Langauge::code(); ?>" dir="<?= Language::direction(); ?>">
```

LuckyCyborg:
@pandory-network 

> set language cookie ucfirst

That I should change, because go to more complex code.

amiroperator:
@LuckyCyborg some bootstrap styles can't switched to right alignment.

LuckyCyborg:
@amiroperator In fact I believe that there should be a **dedicated Template** for.

BTW, look there: https://github.com/zeroxme/bootstrap-3-arabic

Making the **Core\Language** direction aware is a good thing and I support it. 

But for a complete there should be Controllers which are Langauge direction aware, and to switch Templates.

It is possible to do it. Prepare you first a **DefaultArabic** Template on pair with Default one and we talk again about.

Then, remove the changes on Template and please keep only those in Language System.

amiroperator:
@LuckyCyborg You mean we should make a special template for rtl templates ?

LuckyCyborg:
Yep. A complete new one, specially for RTL Bootstrap, on pair with Default Template.

amiroperator:
Would you like me to do that or you will do ?

LuckyCyborg:
If you want RTL, you have to do it. After all, you are one able to appreciate when it works correctly, right?

And, like I said, using new View API, is possible to change the Templates on fly, via Controller's Middleware.

amiroperator:
Should I close this PR and start over ?

LuckyCyborg:
On this PR remove every change on Templates, and leave it for merging. 

I agree with the changes on Core\Language and Config.php.

amiroperator:
Ok

LuckyCyborg:
Your RTL version of Default should be called exactly: **DefaultRtl**

You see after WHY I ask that.

LuckyCyborg:
Or, an alternative idea is to create two Layouts: **default-rtl.php** and **custom-rtl.php** on Default Template.

After I have the material to work and compare, we chose the best solution for RTL way.

amiroperator:
I agree with that idea to create two Layouts

LuckyCyborg:
Considering that the switching is possible only on new View API, I propose to start with those two Layouts.

After, we talk about the complete Template, but for now, do it those two Layouts.

amiroperator:
It is easier to make a folder in default folder named rtl for rtl Layouts 

and in View API we can check if current language direction is set to rtl then load exiting rtl Layouts else  load default Layouts

What do you think ?

LuckyCyborg:
@amiroperator Remained a change into **app/Templates/Default/header.php**

@daveismyname would not like that... :smirk:

amiroperator:
9 commits is too much for a small change !

LuckyCyborg:
@amiroperator That's perfect.

To remember to write those Layouts, first: **default-rtl.php** and **custom-rtl.php** on Default Template. 

Only they could be used under new **View API**.

amiroperator:
Maybe tomorrow I am tired now

LuckyCyborg:
Okay, no problem. 

It is late hour in Iran, right?

amiroperator:
1:35 AM

LuckyCyborg:
Good night, my Iranian friend! 

It is time to shutdown your computer, in my opinion... :smirk:

amiroperator:
Good night.

PR #855, opened 2016-05-20
Fix Fatal error: the name is already in use

amiroperator:
like the title said


LuckyCyborg:
Thanks! PHP7 hit again! :smile:

BUT, the Language Facade is not a **CoreLanguage**, use to alias to **LanguageFacade** eventually.

In fact, I mussing that I would include this fix in my next pull request, for today.

amiroperator:
So should I close it ?

PR #858, opened 2016-05-20
Improve the Language System and the compatibility with PHP7 on Validation\Factory and Validation\Validator

LuckyCyborg:
Like the title say.

To note that also this pull request improve the **Language System**.


amiroperator:
If you could just translate the attributes that would be good. Because e.g username in other languages is understandable:
'username باید حداقل 3 کاراکتر باشد.'

PR #861, opened 2016-05-20
Adjust the Czech Language files, improve the Language System, Core\Config and Validation\Translator

LuckyCyborg:
Like the title say.


geomorillo:
It is not only the cz language

LuckyCyborg:
Then which one is faulting also?

geomorillo:
sorry only cz is faulty

geomorillo:
it think the problem is with the folder there is no cz folder but cs

geomorillo:
oh you fixed it ok

PR #862, opened 2016-05-20
Implementation of a Laravel-esque Request, Input (and Response) support

LuckyCyborg:
This pull request introduce a implementation of a simple but very powerful **Request API**, built in top of **Symfony HTTP Foundation**, and being similar with its counterpart **Laravel 4.x**, of what documentation apply, excluding the **Session** part, which is not (yet) completed / integrated with the Framework.

The objective is to create a fully functional support for HTTP Requests, even for complicated RESTful APIs, in the same time to simplify the input data handling on a daily base.

The usage could be very simple, as in following example, together with the **Validator**:

```php
use Validator;
use Input;

$rules = array(
    'username' => 'required|min:3|max:50|alpha_dash|unique:users',
    'password' => 'required|between:4,30',
    'email'    => 'required|email|max:100|unique:users',
);

// Collect all input data - i.e. $_POST and $_FILES, for a POST Request
$data = Input::all();

$validator = Validator::make($data, $rules);

if ($validator->passes()) {
    // Do something shiny, the input data is perfect.
} else {
    $errors = $validator->errors()->all();

    // Do something nasty, because the data is wrong.
}
```
A more simple example is:
```php
if (Input::has('username')) {
    $username = Input::get('username');

    // Do something shiny with this variable
}
```

But this new Request API is not only about that, for example, you can simply detect a AJAX Request, just like in the actual API:

```php
if(Request::ajax()) {
    // You are set, this is a AJAX Request
}
```
OR see if you have a POST Request:
```php
if(Request::method() == 'POST') {
    // You are set, this is a POST Request
}
```

Finally, there apply almost the entire Illuminate\Request v.4.x API, for example, you can do:
```
if (Request::isJson()) {
    // Get the variable 'result' from the JSON encoded Content
    $result = Request::json('$result');
}
```
OR
```php
// Get the current path info for the request.
$path = Request::path();
```

To note that, while present, it is not possible yet to use the **Session (and Flash)** methods from **Http\Request**, because the **Session\Store** is not full integrated on Nova.

Also, it is not possible yet to use the **Response API**, which while completed, is not yet intergrated full, then it is safely to ignore it for moment.

### DIFF
--- a/app/Config.example.php
+++ b/app/Config.example.php
@@ -166,7 +166,7 @@ Config::set('classAliases', array(
     'TableBuilder'  => '\Helpers\TableBuilder',
     'Tags'          => '\Helpers\Tags',
     'Url'           => '\Helpers\Url',
-    // The Facades
+    // The Support Facades
     'Auth'          => '\Auth\Auth',
     'DB'            => '\Support\Facades\Database',
     'Event'         => '\Support\Facades\Event',
--- a/app/Config.example.php
+++ b/app/Config.example.php
@@ -171,7 +171,9 @@ Config::set('classAliases', array(
     'Auth'          => '\Auth\Auth',
     'DB'            => '\Support\Facades\Database',
     'Event'         => '\Support\Facades\Event',
+    'Input'         => '\Support\Facades\Input',
     'Language'      => '\Support\Facades\Language',
+    'Request'       => '\Support\Facades\Request',
     'Validator'     => '\Support\Facades\Validator',
     // The Legacy Mailer
     'Helpers\PhpMailer\Mail' => '\Helpers\Mailer',
--- a/app/Config.example.php
+++ b/app/Config.example.php
@@ -140,7 +140,6 @@ Config::set('modules', array(
 Config::set('classAliases', array(
     'Config'        => '\Core\Config',
     'Errors'        => '\Core\Error',
-    'Request'       => '\Core\Request',
     'Response'      => '\Core\Response',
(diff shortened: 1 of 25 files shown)


jimgwhit:
Ok @LuckyCyborg I might like this one, how about some good examples of submitting a form and getting the post request with the above pr.  
Look how good Taylor documents fromlaravel docs. Remember, not good doc = confused, good doc = happy and understand.
```
Retrieving An Input Value

$name = Input::get('name');

Retrieving A Default Value If The Input Value Is Absent

$name = Input::get('name', 'Sally');

Determining If An Input Value Is Present

if (Input::has('name'))
{
    //
}

Getting All Input For The Request

$input = Input::all();

Getting Only Some Of The Request Input

$input = Input::only('username', 'password');

$input = Input::except('credit_card');
```

Kingorgg:
You have two 'Request' Aliases, would they not conflict with each other?

LuckyCyborg:
@Kingorgg 

> You have two 'Request' Aliases, would they not conflict with each other?

Thanks for noticing that! 

Sure that double aliasing will generate a conflict, because from now, the old Core\Request should be called **literally via **use**. 

Fixed now.

LuckyCyborg:
@jimgwhit 

> Look how good Taylor documents from laravel docs.

Maybe there **our** Taylor will save the show.... :smirk: 

Personally, like I said, I'm not good to write documentation.

Sorry for that, but I need help in this domain... :sake:

PR #868, opened 2016-05-21
Introduce the ability to handle different Database\Connection instanc…

LuckyCyborg:
…es on Database\Model and Database\ORM\Model

### DIFF
--- a/system/Database/Model.php
+++ b/system/Database/Model.php
@@ -16,6 +16,13 @@ use Helpers\Inflector;
 class Model
 {
     /**
+     * The Database Connection name.
+     *
+     * @var string
+     */
+    protected $connection = 'default';
+
+    /**
      * The database connection instance.
      *
      * @var \Database\Connection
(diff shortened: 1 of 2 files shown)


daveismyname:
usage example?

LuckyCyborg:
@daveismyname 

In the app/Config.php could be defined multiple Database configuration entries. Considering that we add a new entry called 'custom', we have:

```php
namespace App\Models;

use Auth\Model as BaseModel;

use \stdClass;


class Users extends BaseModel
{
    protected $connection = 'custom';

    protected $table = 'users';

    protected $primaryKey = 'id';


    public function __construct()
    {
        parent::__construct();
    }

    ...
}
```

Or the Connection instance name could be given on Model instatiaton:

```php
namespace App\Models;

use Database\Model as BaseModel;

class Articles extends BaseModel
{
    public function __construct($connection = null)
    {
        parent::__construct($connection);
    }

    ...
}
```

```php
$model = new \App\Models\Articles('custom');

// Do something shiny with the Model instance.
```

daveismyname:
Great, thanks.

LuckyCyborg:
@daveismyname No problem.

LuckyCyborg:
@daveismyname Please modify Release Notes the second Model example as I modified there.

For the Database API's Models is mandatory to specify the **table** and **primaryKey**.

daveismyname:
Roger that, updated it.

I'm going to build novacasts using the new database api so that will give me good coverage when doing tutorials with.

LuckyCyborg:
@daveismyname And another small mistake on second Model, the table should be **articles** not **users**. 

Uh oh, I need to start the computer, that tablet have its own mind.

And just ask if you want details about something...

daveismyname:
ha no problem!

PR #869, opened 2016-05-21
This should work the same

geomorillo:
tested it and works ok


LuckyCyborg:
@geomorillo Lighten me, what is the sense of this change?

I leaved as **null** the parameter for future developments, where I have intention to give a Guard name in configuration...

geomorillo:
then ok

PR #872, opened 2016-05-21
Add: Encryption Key generation to the CLI.

Kingorgg:
use 

````php
php nova make:key
```
in your console, to generate a key for the config file.


daveismyname:
Excellent!

PR #873, opened 2016-05-21
Update functions.php

pandory-network:
add function to output formatted and translated message

### DIFF
--- a/system/functions.php
+++ b/system/functions.php
@@ -66,6 +66,23 @@ function __($message, $args = null)
 }
 
 /**
+ * Output the formatted and translated message.
+ *
+ * @param string $message English default message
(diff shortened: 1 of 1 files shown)


LuckyCyborg:
@daveismyname I propose reverting this pull request as being of questionable value. 

What is the sense of introducing **_e()** while Nova is clear to be designed for PHP 5.5 and beyond ?

As example, the actual thing:
```
<?= __('Hello'); ?>
```

Using that function introduced by this PR:
```
<?php  _e('Hello'); ?>
```

Also, to note that the strings on '_e()' function are not detected by the **Language Files Updater**, then those translations will glorious fail, if there wasn't the same string in a '__()' or introduced manually.

PR #881, opened 2016-05-22
Move the Framework to using the new Request / Response APIs

LuckyCyborg:
# Introduction

This pull request move the Framework on using the new **Request / Response APIs** and a Laravel-eaque **Session\Store**.

### Moving... what?

Today, the Nova Framework is structured in two different APIs, one called **Classic APIs**, which offer all features loved by users of previous **SMVC 2.2**, for example: the triplet style rendering or the simple Database Helper, and another one called **New Style APIs**, which offer alternative features, usually more advanced.

Some classes are part of both APIs, for example **Core\VIew**, while another ones are dedicated exclusively to the **New Style APIs**, for example the **QueryBuilder** from **Database API**. 

While big, this pull request affect basically only the **New Style APIs**, literally replacing only **Core\Redirect** and **Core\Response**, both not being part of the **Classic APIs** and already offering an API similar with the ones new introduced.

Then, the basic usage even of those classes, for the end-users who adopted the **New Style APIs**, will be preserved. 

In other hand, almost all components subject of this change are already introduced on previous pull requests, and this one could be seen as putting all together and making the components integration.

### Where is **Core\Request** ?

The **Core\Request** is moved back to **Helpers\Request**, as Legacy Support for preserving the API. Probably all videos refer to it right where it is now.

### Why removing of  **Core\Redirect** and **Core\Response** ?

The **Core\Redirect** and **Core\Response** are removed, their functionality being super-seeded by the new introduced APIs, which are, in principle, compatible with the previous **Response** implementation.

### Where is the Logs directory ?

Because the need to have **yet another directory** where the Session Handler to store the files, is introduced in App a folder called **Storage** which contains **Cache**, **Logs** and **Sessions**.

# API Compatibility

The **Classic API** is preserved with **no changes**, eventually minimal namespace modifications required, in the case of** Helpers\Request**, while the new APIs should use the new **Request, Response and Session Storage APIs**.

But, like was said, the previous  **Response** and **Redirect** have already a similar API, then the basic usage is preserved.

# Notable Changes

#### Using the "use"

The new **Session\Store** should be called usually via its **Facade**, called **Support\Facades\Session**. 

Then, instead of:
```php
use Helpers\Session;
```
you should use:
```php
use Session;
```
OR, if it  is the case, directly the associated Facade:
```php
use Support\Facades\Session;
```

#### Destroying the stored data

The new **Session\Store** make difference between destroying of one Data Item and removing all Data.

```php
// Deleting a specified data item:
Session::forget($key);

// Removing all the stored data.
Session::clear();
```

#### Checking if a key exists in the stored data

The new **Session\Store** use **has()** instead of **exists()** to check if a key exists in the stored data. For example:
```php
if(Session::has($key)) {
    $value = Session::get($key);

    // Do something shiny with the retrieved value
}
```
#### Pulling data from the Session's stored data

Instead of old **Session::pull()**  please use the method **Session::remove()**, as in following example:
```php
// Pull a Message from the Session Store.
$message = Session::remove($name);
```

# Additional Changes

All the Classes which are part of Routing are moved to namespace **\Routing\** and is continued the implementation of the new **Request / Response API** infrastructure.

Notable is also the introducing of a Laravel-like **Encryption\Encrypter** class and moving the **Auth\Auth** Facade to namespace **Support\Facedes**. 

Also to note the moving of **Core\Request** as Legacy Helper to **Helpers\Request**, the entire Framework's structure being moved to the new **Request API**.

# Installation on Sub-Directory Warning

The installation of Nova Framework on a sub-directory in a production server introduce a Security Risk similar to modern Laravel, CodeIgniter or CakePHP: 

The files from **app/Storage/Sessions** contains your saved Sessions data, read: sensitive information in clear-text, and a server configuration failure can make them public available.

**Please, never install Nova Framework in a sub-directory, specially on production servers!**

If you believe that a Server configuration failure will never happen, to know that as the same believed the guys from Facebook, too... :smirk:


jimgwhit:
@LuckyCyborg said
> Small change into Support\Facades\Redirect  

I have been busy, so how does the new redirect work campaired with original? And is a "use" statement needed at the top of the controlled if using redirect in that controller?

LuckyCyborg:
@jimgwhit There is a need of a "use" in top always, not only for Redirect, because App is namespace-d.

To note that the new Redirect API is not active yet.

Also, I do not understand about what "original" Redirect you talk. The existent one, **Core\Redirect*?

jimgwhit:
I am referring to
```
Url::redirect('admin/login');
```

LuckyCyborg:
@jimgwhit 

 > I am referring to
> ```
> Url::redirect('admin/login');
> ```

That's old API, who shutdown the Framework in the middle of nowhere.

The **Routing\Redirector** and its **Facade** offer a Laravel-like API using the new Request / Response APIs.

daveismyname:
wow sooooo many changes!

Put Session::pull in the new class simple, it's a very useful function to use and remove from session in one go why go back to doing it a longer way.

I've just downloaded your repo installed and then got this:

````
An error has occurred. This error has been reported.


Warning: session_write_close(): Session callback expects true/false return value in Unknown on line 0

Warning: session_write_close(): Failed to write session data (user). Please verify that the current setting of session.save_path is correct (/Applications/MAMP/tmp/php) in Unknown on line 0
````

Error:

````
2016-05-23 14:16PM - mcrypt_encrypt(): Key of size 0 not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported in /novalocal/system/Encryption/Encrypter.php on line 93
````

If all sessions are stored to disk how do they get removed to stop them being stored indefinitely. The folder could get very large very quickly.

CLI fails to create key:

````
Fatal error: Call to undefined function Encryption\mcrypt_create_iv() in /novalocal/system/Encryption/Encrypter.php on line 62
````

After putting the key in config:

````
2016-05-23 14:23PM - session_write_close(): Session callback expects true/false return value in Unknown on line 0
````

Welcome controller still have use Helpers\Session;

LuckyCyborg:
@daveismyname 

> Put Session::pull in the new class simple, it's a very useful function to use and remove from session > in one go why go back to doing it a longer way.

Instead o pull there is **remove()**

And thanks for testing, I will look to, immediately I arrive at job.

LuckyCyborg:
@daveismyname Please test it again! :smile:

jimgwhit:
@daveismyname said  
> If all sessions are stored to disk how do they get removed to stop them being stored indefinitely. The folder could get very large very quickly.  

Does the current session helper use file storage, and if so how are they purged from time to time? I agree, session storage needs a purge from time to time.

LuckyCyborg:
@jimgwhit 

> As far as that goes, how does apache handle it with just a raw php file?

There is a Garbage Collector, which delete automatically the stalled files. And yes, even Apache use by default files for Sessions.

See http://php.net/manual/en/class.sessionhandlerinterface.php and other associated pages.

amiroperator:
So many changes. How can we handle these ?? :+1:

LuckyCyborg:
@amiroperator In principle, is only about the Framework infrastructure and the components integration on **Execution Flow**, nothing more.

The principal components integrated there, are already on place from previous pull-requests.

For example, the **Request API**. It is already there.

Another example being the **Response API** which present essentially the same API on basic usage.

Maybe looks scaring this pull request, but, like I said, is more about things for Framework's kitchen.

daveismyname:
Tested all working great! see the chat room for all our comments on testing.

PR #885, opened 2016-05-26
 Improve the Pagination API and its integration with Database API

LuckyCyborg:
This pull request improve the **Pagination API** and its integration with **Database API**.

Also, it introduce a **Facade** for **Pagination API** and ability to configure the current page name, aka the query parameter which specify the Pagination's current Page. Usage:

```php
use Paginator;
use DB;

// Set the custom Page Name
Paginator::setPageName('p');

// Get a Paginator instance from the QueryBuilder.
$paginate = DB::table('posts')->paginate(25);

// Print the Pagination Links.
echo $paginate->links();

// Process the current Page results.
foreach ($paginate->getItems() as $post) {
    echo $post->title .'<br>';
}
```

Also, there is introduced the ability to specify per Model the number of Items returned for a Page on the Pagination\Paginator instance by QueryBuilder, using as following:
```php
namespace App\Models;

use Database\Model;

class Users extends Model
{
    /**
     * The table associated with the Model.
     *
     * @var string
     */
    protected $table = 'users';

    /**
     * The primary key for the Model.
     *
     * @var string
     */
    protected $primaryKey = 'id';

    /**
     * The number of Records to return for pagination.
     *
     * @var int
     */
    protected $perPage = 25;
}
```
That **perPage** setting will be honored, of course, on the Model's associated QueryBuilder instances.

Finally, the **Database\Query\Builder** API is enhanced for a better Pagination integration.


LuckyCyborg:
@daveismyname Please merge this pull request, its objectives are arrived to.

PR #899, opened 2016-05-29
Introduce in the Database API the support for using PostgresSQL and SQLite Databases

LuckyCyborg:
This pull request introduce in the **Database API** the support for using **PostgresSQL** and **SQLite** Databases also, behind of the **MySQL** ones.

**To note that this pull request does NOT introduce any API changes of breaks.**

### DIFF
--- a/app/Config/Database.php
+++ b/app/Config/Database.php
@@ -24,4 +24,11 @@ Config::set('database', array(
         'charset'   => 'utf8',
         'collation' => 'utf8_general_ci',
     ),
+    'custom' => array(
+        'driver'    => 'sqlite',
+        'database'  => APPDIR .'Storage' .DS .'database.sqlite',
+        'prefix'    => PREFIX,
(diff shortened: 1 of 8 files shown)


jimgwhit:
@LuckyCyborg on the othe thing, OK.  
After pr #898 is there any change on how to use the new api, in other words do I need to change default to custom, or will either work out of the box now?  
Like you showed the other night with the 
~~~
use DB;
~~~
Is that still all I need?  Or is this default and the helper one now have to be assigned custom?

LuckyCyborg:
@jimgwhit 

You have to specify the used configuration **only** when you want other instances and configurations than the **default** ones, as in example:
```php
// Get the default Database instance
$db = Database::get();

// Get a custom configured Database instance
$customDb = Database::get('custom');
```
To note that the **Helpers\Database** use now the **app/Config/Database.php** for defining the connections, just like **Database API**

PR #900, opened 2016-05-29
A simple but very handily improvement of Core\Controller

LuckyCyborg:
This pull request introduce a simple but very handily improvement of **Core\Controller**, making it capable to transparently call a View instance, if the developers accept to use a implicit View for the Controller's method.

 As a example of the usage for the Controller **App\Controllers\Welcome**:
```php
public function subPage()
{
    $this->title($this->trans('subpageText'));

    return $this->getView()
        ->withWelcomeMessage($this->trans('subpageMessage'));
}
```

Another example, for **App\Modules\Users\Controllers\Users**:
```php
    public function login()
    {
        $error = Session::remove('error', array());

        $this->title(__d('users', 'User Login'));

        return $this->getView()
            ->with('csrfToken', Session::token())
            ->with('error', $error);
    }
```

How it is calculated the implicit View path?

It is the calculated from the Controller complete name (with namespace), replacing **'\Controllers\'** with **'\View\'** and add as the last element, the titled **method**. For example:

**App\Controllers\Welcome@subPage** -> the View: **'Welcome/SubPage'** 

**App\Modules\Users\Controllers\Users@login** -> the View: **'Users/Login'**, the module: **'Users'**

Finally, there is added a new method, called **title()** which is practically a shortcut to adding the title to View's shared data:
```php
    /**
     * @param  string $title
     *
     * @return \Core\Controller
     */
    protected function title($title)
    {
        View::share('title', $title);
    }
```

### DIFF
--- a/app/Modules/Users/Views/Users/Login.php
+++ b/app/Modules/Users/Views/Users/Login.php
@@ -22,6 +22,10 @@
                 <div class="form-group">
                     <p><input type="password" name="password" id="password" class="form-control input-lg col-xs-12 col-sm-12 col-md-12" placeholder="<?= __d('users', 'Password'); ?>"><br><br></p>
                 </div>
+                <div class="form-group">
+                    <p><input name="remember" type="checkbox"> <?= __d('users', 'Remember me'); ?></p>
+                </div>
+                <hr>
                 <div class="row" style="margin-top: 22px;">
                     <div class="col-xs-6 col-sm-6 col-md-6">
                         <input type="submit" name="submit" class="btn btn-success col-sm-8" value="<?= __d('users', 'Login'); ?>">
--- a/system/Core/BaseView.php
+++ b/system/Core/BaseView.php
@@ -311,7 +311,7 @@ abstract class BaseView implements ArrayAccess
      * Magic Method for handling dynamic functions.
      *
      * @param  string  $method
-     * @param  array   $parames
+     * @param  array   $params
      * @return \Core\BaseView|static|void
      *
      * @throws \BadMethodCallException
--- a/system/Core/Controller.php
+++ b/system/Core/Controller.php
@@ -205,12 +205,12 @@ abstract class Controller
     {
         View::share('title', $title);
     }
-    
+
     /**
      * Return a implicit View instance.
      * @return \Core\View
      */
-    protected function view()
+    protected function makeView()
     {
         // Setup the Controller's properties.
         $className = get_class($this);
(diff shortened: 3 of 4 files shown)


daveismyname:
What's the purpose of adding title when it's part of the view?

what does getView do? it does not accept any params.

LuckyCyborg:
@daveismyname 

**title()** is a shotrcut for adding a shared variable to Views. Those two are equivalent:
```php
        $this->title(__d('users', 'User Login'));
```
```php
        View::share('title', __d('users', 'User Login'));
```

**getView()** return a View instance using a precise calculated View path, according with the called Controller. Those two are equivalent:
```php
        return $this->getView()
            ->with('csrfToken', Session::token())
            ->with('error', $error);
```
```php
        return View::make('Users/Login', 'Users')
            ->shares('title', __d('users', 'User Login'))
            ->with('csrfToken', Session::token())
            ->with('error', $error);
```

daveismyname:
so for it to work the view **must** be the same name as the method, in terms of speed is this slower then doing View::make('Welcome/Sub') since it has to calculate the path.

What's the benefit are methods being added just to do it an alternative way?

LuckyCyborg:
@daveismyname 

> so for it to work the view **must** be the same name as the method, in terms of speed is this slower then doing View::make('Welcome/Sub') since it has to calculate the path.

Come on! You know well that **99%** of time the framework spend on retrieving data from database, processing it and eventually inserting something back on database. :smirk:

The speed difference between the two variant to make a View instance will be insignificant!

> What's the benefit are methods being added just to do it an alternative way?

The main benefit is that myself, personally, I have intention to use it. I consider it very handily, even is a **must** to respect some rules for Views path. I already use those rules.

LuckyCyborg:
@daveismyname There is a example from a CRM written on NovaIgniter:
```php
    public function create()
    {
        $centers = Center::findAll();

        return View::make()
            ->shares('title', __('Create Student'))
            ->with('csrfToken', Csrf::makeToken())
            ->with('centers', $centers);
    }
```
That use a Controller aware Views.

BUT, instead to replace manually the things, on every Controller, I could do:
```php
    public function create()
    {
        $centers = Center::findAll();

        return $this->getView()
            ->shares('title', __('Create Student'))
            ->with('csrfToken', Csrf::makeToken())
            ->with('centers', $centers);
    }
```

jimgwhit:
@LuckyCyborg is doing the implicit call completely optional? And I really hope the answer is yes. Gone too far now to start changing things.

LuckyCyborg:
Yes. 

You have to choose what you want, a View instance using implicit parameters, or the normal way, making the View instance yourself.

PR #934, opened 2016-06-05
Improve Demo, Update persian language and Make RTL Layouts Up-to-date

amiroperator:


vahidalvandi:
mr amir for RTL version of bootstrap i have solution in core bootstrap ,,,

LuckyCyborg:
@vahidalvandi 

> i have solution in core bootstrap ,,,

We ship now the **Bootstrap 3** from a external composer package, then no one to dare to hack it. :smiling_imp: 

If you can live with your RTL changes on top of it, as it is, that's OK.

vahidalvandi:
bt 3.5  rtl version like ltr with same size but use other package side original increase size and this not Good in speed loading...

PR #956, opened 2016-06-07
Implementation of a Multilingual support on Routing, Users Registration on Users module and the Database support for the Session Store and Framework configuration.

LuckyCyborg:
This pull request introduce a **Multilingual support** on **Routing**, the **Users Registration**, with E-mail activation, on the Users module; and **the Database support for configuration**, accessible usually via its **Facade**.

To note that the new Configuration engine, living in the namespace **\Config**, is **NOT** a replacement for the **Core\Config**.

**IMPORTANT:** The configuration of the Framework, on the associated config files from App, should be done via **Core\Config**, as was always.

The new Configuration Engine will act more like a extension for **Core\Config**, to support the **transparently saving and reading the configuration to Database**, very useful when you have configuration pages in your application, for example giving a configuration page for **Emailer**, site name, etc.

For **enabling the Database support** on **Config API** you should setup according the define **APPCONFIG_STORE**, located on `app/Config/app.php`, to `database`, and requiring a dedicated table imported from scripts, while on the default mode `files` offering only a support to reading the App configuration, aka `Config::get()`.

In other hand, this pull request introduce the class **Session\DatabaseSessionHandler**, which permit to use for **Session API** the Database as backend. This new facility needs, of course, a associated database table, as imported from SQL scripts.

Finally, to note that **there are no API changes or breaks**.


app/Config/App.php
```diff
@@ -14,8 +14,11 @@
  * The Application configuration.
  */
 Config::set('app', array(
-    'siteUrl'  => SITEURL,
-    'siteName' => SITETITLE,
+    'siteUrl'      => SITEURL,
+    'siteName'     => SITETITLE,
+
+    // Wheter or not is the Application multilingual.
```
Kingorgg: "Enable multilingual support."

PR #958, opened 2016-06-08
Proposal to deal with multilanguage

avenirer:
Hello,

I've seen how the framework deals with "languages", but I think the module was poorly implemented. Ok, I know that you can change the language by accessing the language controller and the change() method. But the module doesn't seem complete. What I mean by this is that the language slug ("en", "fr", "ro") should find itself inside the url, because if not, I think the search engines will consider that the content served on the same url is different, as the same urls will serve "same content" but in different languages. I sure hope you understand what I mean here...

Now, in order to prevent this from happening we should have the language slug appended to the url, if the language is not the "default" one. I tried to deal with this inside the Url helper, but it doesn't seem to work, as the defined routes would bypass the parsing of the url.

The only viable solution was inside the Router.php.

Now, if the framework has more than one language defined, the dispatch() method will look if the slug is inside the url, if the slug is passed inside the url, then the language is set inside a session and a cookie.

If inside the url there is no language defined, the dispatch() method will look for the language inside the cookie and the session and if it finds out that the language is different than the default language, then it will redirect to an url that has the language slug appended.

This way of dealing with the url would also permit us to simply eliminate the language controller, as everything regarding the language would be dealt with inside the router.

Hoping that this would be aproved...

### DIFF
--- a/system/Routing/Router.php
+++ b/system/Routing/Router.php
@@ -13,12 +13,14 @@ use Helpers\Inflector;
 use Helpers\Url;
 use Routing\BaseRouter;
 use Routing\Route;
+use Helpers\Cookie
 
 use Symfony\Component\HttpFoundation\Response as SymfonyResponse;
 
 use App;
 use Response;
 use Request;
+use Session;
 
 
 /**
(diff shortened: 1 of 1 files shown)


LuckyCyborg:
@avenirer First of all, as we discussed on team before, is better to leave to end-user that multi-language content, which make sense when you make a CMS on top of Nova.

Multi-language content is totally a dead weight for an API or full private Portal, i.e. a CRM.

In other hand, the Nova's Routing permit adding new patterns in its configuration, and I even added an optional **:lang** one in my PR.

Finally, there is the Middleware, which can do some pre-processing.

In fact is spectacular to implement the multi-language content without introducing multi-lingual processing on Routing.

LuckyCyborg:
BTW, mixing the Classic Style and new Style APIs is not a good thing.

Bear in mind that any Helpers which have a New Style API equivalent are preserved there as **legacy support**. 

Also, that internally Nova use the New Style ones, then using `Helpers\Cookie` in Routing is not the luckiest idea, specially combined with the new **Session API**.

Finally, playing with the DIR define is a thing which I do not like.

If you are interested in this multi-language content, try to find a more acceptable solution, with minimal intervention on Routing.

LuckyCyborg:
```php
Router::any('(:lang)(/(:all))', 'App\Controllers\Pages@showPage');
```
On Controller:
```php
public function showPage($lang, $path = null) 
{
    // Some processing there
}
```
That does the trick without touching the routing.

LuckyCyborg:
@avenirer Then, I added support for multi-language in my https://github.com/nova-framework/framework/pull/956 , in the way which I consider right.

Like you guess, the multi-language is not a simple thing, then there is a option on **app/Config/App.php** and I moved the default Welcome demo to multi-language.

The Routing will detect the localized URI, while the Middleware will apply the localization. Also to note that no DIR using for you, only the `site_url()` to be used, and for translating strings is better to use the new **Language API**, the legacy one being more difficult on this case.

PR #960, opened 2016-06-09
Improve Auth\Reminders and add the infrastructure of the Service Providers

LuckyCyborg:
Like the title say.

NO API changes included!

### DIFF
--- a/app/Config/App.php
+++ b/app/Config/App.php
@@ -61,14 +61,17 @@ Config::set('app', array(
      * The registered Service Providers.
      */
     'providers' => array(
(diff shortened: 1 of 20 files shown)


daveismyname:
this for internal use or for end user to leverage? if so documentation will be required.

LuckyCyborg:
@daveismyname Good point! And I know that, worry not. :smirk: 

Basically in this PR is only about **Internal API**, not being exposed to end-user and for optimization and; making further extensible the Framework.

LuckyCyborg:
@daveismyname Thanks!

PR #973, opened 2016-06-16
Updated version number ...

Kingorgg:
in system/Foundation/Application

### DIFF
--- a/system/Foundation/Application.php
+++ b/system/Foundation/Application.php
@@ -44,7 +44,7 @@ class Application extends Container implements HttpKernelInterface, TerminableIn
      *
      * @var string
      */
-    const VERSION = '3.55.2';
+    const VERSION = '3.56.0';
 
     /**
      * Indicates if the application has "booted".


daveismyname:
ah, yes this is updated in the release.

LuckyCyborg:
@Kingorgg and @daveismyname Sorry for mess, but because it is a **const** I can't update on-fly, then we should manually adjust too.

daveismyname:
Since you made me away I've been updating it in the releases must remember to do the source files too.

LuckyCyborg:
@daveismyname The good part will be when we arrive with system on Composer package. Where this const will be hardcoded on package, while VERSION.txt will indicate the **nova-framework/app** too.

LuckyCyborg:
@daveismyname BTW, you will hugely simplify the tasks of your end-users if you move the system to Composer package...

LuckyCyborg:
@daveismyname Sadly, to move to Composer package with the system, while extremely simple to do, need a common namespace for it. 

Was not a luckiest decision to have the system components without a common namespace...

PR #983, opened 2016-06-19
Move the Profiler and ReCaptcha configuration to their own files and simplify the App configuration

LuckyCyborg:
### Rationale

In the today Nova, where almost all configuration options are moved to their own files, make no sense to still force the end user to rename the **Config.example.php**, containing just a hand of options.

As final step, the Profiler and ReCaptcha configuration is moved to their own files and **app/Boot/Start.php** load directly the configuration files located on **app/Config** directory.

To note that the **Boot Stage** will load first the file **app/Config/Constants.php**, then every other configuration file.


daveismyname:
2 problems:

1) if composer has not been initiated instead of receiving useful instructions, the user is presented with errors.
2) a database connection is being forced, it should be optional.

LuckyCyborg:
@daveismyname 

>  if composer has not been initiated instead of receiving useful instructions, the user is presented with errors.

Sure, a error will appear if Composer will not be run, as `vendor/autoload.php not found`. Yet, I believe that our end-users are NOT retards, they will read the **README**, also the documentation,  and know how to use Google, where probably will figure immediately what they have to do. 

> 2) a database connection is being forced, it should be optional.

Where you seen you a Database required? That patch move around configuration and defines, BUT do introduce nothing new.

daveismyname:
Come on! just cos it can be figured out does not mean basic instructions that have been there since day 1 should be removed!

as soon as I did a composer install I get:

````
SQLSTATE[HY000] [1045] Access denied for user 'nova'@'localhost' (using password: YES)
````

LuckyCyborg:
@daveismyname Maybe long time ago, SMVC was for absolute PHP beginners, but the today Nova is a mini-Framework, where a person like this probably will have much bigger problems that that error message. :smirk:

And we are the single Framework giving instructions live, if the composer wasn't run.

You consider it really necessary?

Talking about Database, was the Profiler which asked for, being instructed as. Fixed.

daveismyname:
In my mind it still is, it's not about if someone can figure it out it's a bad experience that is not needed when a simple if statement gives the information required.

it's just this:

````
if (!file_exists(ROOTDIR .'vendor/autoload.php')) {
    echo "<h1>Please install via composer.json</h1>";
    echo "<p>Install Composer instructions: <a href='https://getcomposer.org/doc/00-intro.md#globally'>https://getcomposer.org/doc/00-intro.md#globally</a></p>";
    echo "<p>Once composer is installed, navigate to the working directory in your terminal/command prompt and enter 'composer install'</p>";
    exit;
}
````

I've cleared by copy and updated with the latest, the database problem is resolved.

LuckyCyborg:
@daveismyname Added back the message, BUT I really believe that it is over-nurturing.

daveismyname:
I know you do but the small details are important in my opinion.

PR #999, opened 2016-07-14
Fixing Csrf with custom name, wasn't returning the right value.

tomvlk:


tomvlk:
Btw, the way of generation is not really random-safe.

LuckyCyborg:
Well, I agree that **Helpers\Csrf** have a questionable behavior...

**Also, after all, why we need multiple CSRF tokens?**

Bottom line, to note that this Helper exists on Nova 3 just a glorified legacy compatibility with SMVC (2.2), as many other things, and it is nuked on (future) Nova 4.

PR #1019, opened 2016-07-28
Improve the Assets Serving via Routing

LuckyCyborg:
This pull request improve the Assets Serving via Routing, specially the Caching Control.

To note that while the Routing send now the proper Caching headers and it is able to respond proper with 304 (Not Modified), the real results may vary, depending on web-server configuration and may be need to disable the Legacy Session support, via commenting `LegacySession::init();` on **app/Boot/Global.php** and completely not using **Helpers\Session** and associated ones, ie. **Helpers\Csrf**.

Long story short: the Browser side Caching of the Assets served via Routing is subject of **a advanced fine tuning** of the framework for the target web-server. And that may include stopping using the shiny SMVC Legacy support and the cool development on localhost.

Finally, **No API Changes** included there.


geomorillo:
why delete all command files? are not used anymore?

LuckyCyborg:
@geomorillo 

> why delete all command files? are not used anymore?

The two removed files are **not** commands for **nova** console utility, they are the Console Kernel, which on Nova 4 are the bases of **The Forge**.

How the Nova 4 is in an advanced stage, continuing to push features to Nova 3 make no sense, it will receive only bug-fixes or small improvements like on this pull-request.

PR #1044, opened 2016-08-04
Implementation of Before/After Filters for Routes

LuckyCyborg:
This pull request add a complete **Middleware** stack for Routes, similar with Nova 4, now being available the both **Before and After Filters** for them.

Visible changes and need-to-fix: you should rename on your Routes definition the "filters" keys as "before", both more appropriately with their inner work (as from always they are Before Filters), but also because now being available the "after" Filters, then for more naming consistency.

The new After Filters receive a third parameter, which is the Response instance obtained from Before Filters or Router Action execution.

No API changes there, just some new features and a small (non-critical) change on Routes definition syntax.


daveismyname:
noticed the language changer isn't changing on index and subpage downloaded the files from your repo running from built in server.

Language changer is working for the modules though. I'm out this evening so won't be able to do anything more till later.

daveismyname:
can you explain this please:

{slug:.*:?}

I get slug is the name but what is :.*:? that's a combination of 5 characters

LuckyCyborg:
**Language Changer** works perfect as always. The bad thing is that beyond hundred releases we are still stuck on **App\Controllers\Welcome** on that SMVC Compat Language Support which no one care today. 

If you lurk daily on chat, the main questions are today about ORM and how **View::make()** or **getView()** works, then this Welcome Controller was ported to the native Language and Views API. Let's grown up! Our gran-fathers made their History, but there is no need for us to try to use today the Old English, right? Same there, let's leave SMVC rest in peace!

**{slug:.*:?}**, is a Named Parameter called "slug", acting like an "(:all)", as given by second element: ".*", which is custom regex, then the last part say that it is also a **optional parameter**.

daveismyname:
you may not like it but it was agreed to support both ways otherwise it may as well just have been nova.

Not everyone that uses the framework will be active on the chat rooms, that's normal for any project.

Concentrate on Nova 4 I've given you free rain on that, but Nova 3 I want to keep it so it supports both styles as it's done for quite some time.

The language changes is not working for me, I've uploaded a quick video to demonstrate I'm telling the truth.

https://www.youtube.com/watch?v=WD1IhxEFG9g&feature=em-upload_owner

I don't use languages but others do.

LuckyCyborg:
Both "ways/styles" on Routing definition and all other know things about Nova 3 are full supported. 

Long story short: no API changes are introduced, excluding maybe renaming keyword "filters" to "before" on Routes definition.

Also there is no reason for Language Changer to not work. Fact that there are no visible changes on changing the languages is because the strings from associated "messages.php" are not translated (yet). Basically you hit a feature of native Nova 3's Language System, to give you the original strings if there is no translations.

Why there are no translations? Because they was on the ancient Welcome.php file, which was always Compat Support. If one of your lieutenants will be kind to made the required translations, everything will be OK. 

I gues that Danish, Persian and Russian will come up quickly, so nothing to worry. :smirk:

daveismyname:
In that case it would have made sense to have the translations for the welcome page like in the past or remove the language changer as a default view file.

As you know I'm only been promoting the newer style API codes as I agree they make more sense so talk like this is not needed.

> Let's grown up! Our grand-grand-fathers made their History, but there is no need for us to try to use today the Old English, right? Same there, let's leave SMVC rest in peace!

Lets leave this discussion here, you've said your point, I've said mine.

Recently we keep clashing over things that shouldn't even need talking about.

LuckyCyborg:
Thanks!

PR #1050, opened 2016-08-06
Update russian messages.php

XTRO123:


LuckyCyborg:
To note that are Language files on app and system folders.

PR #1053, opened 2016-08-06
update russian messages.php

XTRO123:
I do not understand why not being tested


LuckyCyborg:
Looking on your branch, you are **18 commits** behind the master.

I suggest you to fetch and merge from remote master (Nova)

PR #1077, opened 2016-08-17
Improve the Events Dispatcher, add the Bootstrap support to Modules and improve the CRON support

LuckyCyborg:
This pull request introduce several improvements.

First of all, the **Event Dispatcher** is improved, now using the Application's **IoC Container** to generate the class Listener instances, permitting to use any registered Service as Event Listener. The flip side of this changes is the fact that now the methods on class based Event Listeners **should not be static**, but rather normal methods.

Secondly, is introduced a **Bootstrap support for Application and Modules**, permitting the execution of diverse operations on Boot Stage.

Thirdly, the **CRON** support is heavily improved, now shipping a **Cron Service** with the associated Facade, and becoming Events driven. The Cron Module is removed while the CRON web interface is moved to the **System** Module.

To note that now is possible to create a console command for CRON, then to be executed also on console, specially useful for very long tasks execution.


daveismyname:
ready for merging?

LuckyCyborg:
@daveismyname Yes. :+1:

PR #1083, opened 2016-08-20
Implementation of View Rendering Engines

LuckyCyborg:
This pull request introduce a new concept on Nova's Rendering: the **Rendering Engines**, basically isolating the rendering part in a reusable component. 

While by default, the single **Rendering Engine** available being the PHP one, offering the default Nova's Rendering style, the end-user can implement custom ones, for example on top of **TWIG** or even porting the **Laravel's Blade**, and to use them transparently.

Any custom **Rendering Engine** should implement **View\Engines\EngineInterface** and to offer the `render()` method, according with it. For example:
```php
use View\Engines\EngineInterface;

class TwigEngine implements EngineInterface
{
    /**
     * Create a new TwigEngine instance.
     *
     * @return void
     */
    public function __construct()
    {
        //
    }

    /**
     * Get the evaluated contents of the View.
     *
     * @param  string  $path
     * @param  array   $data
     * @return string
     */
    public function render($path, array $data = array())
    {
        // Return some TWIG goodness there.
    }
}
```

To make the new Rendering Engine active, it should be registered, on View file extension base, on App's or one of Module's **Bootstrap**, with a command similar to:
```php
use App\Extensions\View\TwigEngine;

View::addExtension('twig', 'Twig', function() {
    return new TwigEngine();
});
```

To note that, while the custom Rendering Engine will be activated (automatically) only when a View file with matching extension will be asked, its instance is shared stored, similar with a **Singleton**, that's why is denied to store variable on this class instance and/or use static properties or methods.

This pull request introduce **no APIs break**, excluding the lack of ability to access on the View files (as for the legacy compatibility to SMVC) of meta-**$data** array, which contains all passed variables.


geomorillo:
nice

PR #1105, opened 2016-09-08
Update ReCaptcha.php

ElvinYusifli:


LuckyCyborg:
Thanks for your effort, and for raising my attention to the **Helpers\ReCaptcha**, but I will ask you to be kind to accept as us to close this pull request.

That because this Helper is part of the New Style APIs and it should not depend on a legacy thing. 

So, the proper way is it to use **Support\Facades\Request** instead of **Helpers\Request**.

ElvinYusifli:
Request class not found Input function. Please enable ReCaptcha check.

LuckyCyborg:
I will make changes for this ReCAPTCHA Helper on my current pull-request, then I ask for a bit of patience.

PR #1106, opened 2016-09-10
PHP's default locale and Carbon's locale setting

UngarMax:
PHP and Carbon will now use the Config's locale.


LuckyCyborg:
Nope. That's no good for Nova is capable to dynamically change the Language, so nope.

Look on system/Language to understand how that works.

LuckyCyborg:
I will close that PR, because I added a proper (but inspired by your code) locale setup on https://github.com/nova-framework/framework/pull/1104/commits/c07b1d97e7e068d472873f87c8b5f484c788551a

Thank you.

PR #1131, opened 2016-10-24
Improve the HTTP Exceptions handling

LuckyCyborg:
This pull request improve **the HTTP Exceptions handling**, simplifying and unifying the logic, making possible to expose the entire Error Pages generation to the end-user, them making possible their customization as Template, Layout, etc.

As a collateral consequence, the `Response::error()` was removed, not being useful anymore.

As usage, basically we will go back to roots, using for aborting of the processing, the `App::abort()` which exists there since ages. For example, if you want to go **Error 404** in your Controller's Action, you can use:
```php
App::abort(404, 'Page not found');
```
Same way for going i.e. **Error 400**
```php
App::abort(400, 'Bad Request');
```

### DIFF
--- a/app/Boot/Global.php
+++ b/app/Boot/Global.php
@@ -10,42 +10,45 @@ Log::useFiles(storage_path() .DS .'Logs' .DS .'error.log');
 // Application Error Handler
 //--------------------------------------------------------------------------
 
+// The standard handling of the Exceptions.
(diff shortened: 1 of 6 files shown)


Kingorgg:
Nice :+1:

LuckyCyborg:
@Kingorgg Thanks!

PR #1250, opened 2016-12-29
Update Comments for App\Config files.

Kingorgg:


LuckyCyborg:
@Kingorgg Thanks, Daniel!

Also, the same should be done for the branch **4.0**, in my opinion. 👍

PR #1374, opened 2017-01-23
Danish Language Update

DonWagner:


LuckyCyborg:
Thanks, Padre Wagner! 👍 

When I see something similar on **4.0** branch? 💃

PR #1375, opened 2017-01-24
Danish language update

DonWagner:


LuckyCyborg:
Padre Wagner, thanks you for wanting **to update the Nova 3 to Nova 4** code, BUT I believe that not only me, but also Dave and Daniel will strongly disagree with this idea! 🗡 

So, NOPE.

Please hit your PR to the right branch on our repository!

PR #1588, opened 2017-02-28
Danish language update

DonWagner:


LuckyCyborg:
Thanks, Wagner! :+1:

See also the bare app, if you are kind...

PR #1707, opened 2017-05-16
Update .htaccess

sitchi:
change directory public to webroot.


LuckyCyborg:
Nope! Should stay as it is. 😈 

Long story short, this is a honey trap for knowing about those who try to install Nova naked inside webroot, which is strongly **NOT** recommended.

PR #1771, opened 2017-07-12
Fix htaccess after commit 9e19cdb9aaf0e6c59652bec7b60fb562898f7085 ( Rename public folder as webroot )

Nightprince:


LuckyCyborg:
A **Final Nope**, this is for real a honeytrap for those who consider themselves smarter than the framework developers and try to install Nova naked in the web.

IF you install Nova as is supposed to be, this .htaccess file would go above webroot and will have zero influence. 

Again, Nova is NOT designed to be installed naked on web, it is NOT like Wordpress. 

It is supposed to be installed above document root, and the site webroot to point to `webroot` directory. Just like a submarine with only the periscope out of water.

PR #1778, opened 2017-07-18
small fix for children items menu

XTRO123:


LuckyCyborg:
Hi, Victor!

First, the patch is wrong. Eventually here should be `$child['icon']`, but the usage of `fa-circle-o` was intentionally in the second level of menu (children).

IF you change that, please give me a fallback to default, as is supposed the child items to not have individual icons defined.

XTRO123:
Hi. Oh, yes. Eventually here should be $child['icon']. I need to remember how to cancel PR.

LuckyCyborg:
I can close your PR, and refuse it, if you want. 

Anyway, looks that in the final, it do no changes...

PR #2004, opened 2018-01-06
3.x

XTRO123:


LuckyCyborg:
Thanks, Victor! 👍

XTRO123:
Also we need to fix the Model (App\Modules\Users\Models\User) in this module.  ⚙️

LuckyCyborg:
Well, this Model from the Chat module is not (yet) used, in fact.

But, of course better it should be fixed. 👍

PR #2067, opened 2018-04-29
4.0

mikweb:
use BASEPATH instead of STORAGE_PATH in User profile image.


LuckyCyborg:
> use BASEPATH instead of STORAGE_PATH in User profile image.

Thanks, I got the idea. And I will fix this in the next release. :+1: 

Sadly, I will be forced to reject this pull request, because it has errors.

PR #2119, opened 2019-04-27
Update Registrar.php

l2wok:
not use Carbon

### DIFF
--- a/modules/Platform/Controllers/Registrar.php
+++ b/modules/Platform/Controllers/Registrar.php
@@ -18,6 +18,7 @@ use Nova\Support\Facades\Input;
 use Nova\Support\Facades\Redirect;
 use Nova\Support\Facades\Validator;
 use Nova\Support\Str;
+use Carbon\Carbon;
 
 use Shared\Support\ReCaptcha;
 
@@ -246,7 +247,8 @@ class Registrar extends BaseController
         }
 
         $validity = Config::get('platform::tokens.activation.validity', 60); // In minutes.
-
+        
+        // FatalThrowableError (E_ERROR) = Class 'Modules\Platform\Controllers\Carbon' not found
         $oldest = Carbon::parse('-' .$validity .' minutes');
 
         try {


LuckyCyborg:
Good catch. Thank you!

l2wok:
I think its primitive for you, but very strong for me... Small question, if
you not very busy.
i have example(oauth), like this:


> if (!isset($_GET['code'])) {
> // Fetch the authorization URL from the provider, and store state in
> session
> $authorizationUrl = $provider->getAuthorizationUrl();
> $_SESSION['oauth2state'] = $provider->getState();


But i not see in v4 framework any *Session helper*.
I use only mini_v2 framework before and have small trouble for this moment.

Can you help me. What way i can take for this solve and store oauth2state for
latter use?

Fear my english very bad. If i not wrong - you can read russian lang and
maybe you best read this:
Как мне лучше сохранить state, без использования сессий для дальнейшего
использования. Нормально ли будет передать его в cookies? Или что вы
посоветуете для реализации twitch || google oauth?

сб, 27 апр. 2019 г. в 10:45, LuckyCyborg <notifications@github.com>:

LuckyCyborg:
Nova4 does not use the PHP Sessions, but its own Sessions Service, which is very powerful. For example, it uses encrypted cookies.

The usage of Sessions Service is usually via its Facade, for example you can do
```php
uses Nova\Support\Facades\Session;

Session::set('oauth2state', $provider->getState());

// Later you can retrieve this stored value with:
$auth2state = Session::get('oauth2state');
```
Speaking about my knowledge of Russian language, unfortunately I do not speak Russian - my knowledge is limited to ability to read the Russian alphabet, nothing more. 

However, there are users of Nova who speaks natively the Russian language, one of them being @XTRO123 (Victor Efremov) who's also the author of Russian translations from the Nova repositories.

For further questions, I strongly recommend you to use either the Issues or the Users Chat, because using a pull-request's comments for making questions is not exactly the best way - as may be easily overlooked,
 
Issues: https://github.com/nova-framework/framework/issues
Nova Users Chat: https://gitter.im/nova-framework/framework/novausers

I would like to note that Nova community being formed from people born and living on many places from Earth, we have the convention to use the English language on communication, also I could say that the "native" language of Nova is English, because the project leader and its creator, @daveismyname is from United Kindom.

